PR #2, opened 2023-07-31
Add windows soundsystem

Stommm:
Add basic functionality to the playAt method only can play sounds in 2D.

### DIFF
--- a/platforms/windows/SoundSystem_windows.cpp
+++ b/platforms/windows/SoundSystem_windows.cpp
@@ -6,43 +6,14 @@
 	SPDX-License-Identifier: BSD-1-Clause
  ********************************************************************/
 
+
+#include "../../source/Base/Utils.hpp"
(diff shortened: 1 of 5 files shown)


iProgramMC:
Nice! `SoundSystemSL` doesn't do 3D sound either in the original, it only uses the volume.

iProgramMC:
Also, why the hidden window? Include `"Utils.hpp"` and call `GetHWND()`.

Stommm:
Oh, i didnt know that there was a GetHWND function

iProgramMC:
How do you think I create the main window :)

Stommm:
haha, i have not explored the source code that much i just wanted to add sound the instant i realized there was none

iProgramMC:
Actually I think the channels thing is incorrect. Maybe it means "size of sample in bytes" instead.

PR #4, opened 2023-07-31
Create screenshots folder if it doesn't exist

Vruk11:
When taking a screenshot the windows implementation will now check for a folder to save to and create it if it does not exist

g_pApp->field_D58 is now g_pApp->m_externalStorageDir


iProgramMC:
Looks good to me :)

PR #5, opened 2023-07-31
Add enhancement to support scroll wheel for inventory

2008Choco:
It was rather tedious having to switch inventory slots using only the number keys so this enhancement enables the mouse wheel to scroll between slots.


iProgramMC:
Looks good 👍

2008Choco:
My mistake, I should have probably updated the README as well. Would you like me to open another PR?

iProgramMC:
Why?

PR #8, opened 2023-07-31
Surround tickMouse() call in ORIGINAL_CODE preprocessor check

2008Choco:
The call to `tickMouse()` in `Minecraft.cpp` seems to be missing a preprocessor check for `ORIGINAL_CODE`. Despite the fact that this definition doesn't still work yet, this will save some headaches later :)


iProgramMC:
Thanks, didn't notice, haha

PR #9, opened 2023-07-31
Fix options.txt overriding in-game changed settings at runtime

2008Choco:
With the introduction of the `options.txt` support for Windows platforms, changing an option in the options menu while in-game would get overridden each time by the value set in the options.txt file.

This PR adds an overloaded method to `Platform` that accepts a `bool` indicating whether or not the options file should be read. This is then called in `OptionsScreen#remove()` which is responsible for updating the in-game options. Other calls to `getOptionStrings()` remain untouched as `true` is the default value.


iProgramMC:
Well, that's weird, and it's definitely not the way to go about it. Just put the `reloadOptions()` call behind an `#ifdef ORIGINAL_CODE`

2008Choco:
Yeah I suppose that's another way to go about it. A lot simpler than this changeset :)

PR #13, opened 2023-08-01
Platform Definitions

Vruk11:
Made a platform definitions header file for any sort of platform dependent code switching needed.

Made the Sound Engine use the sound system defined in the platform definitions file.


source/Sound/SoundEngine.hpp
```diff
@@ -15,13 +15,14 @@
...
 // Platform specific type for the sound system.
 
-#ifdef _WIN32
-#include "../../platforms/windows/SoundSystem_windows.hpp"
-#define SOUND_SYSTEM_TYPE SoundSystemWindows
-#else
+#ifdef ORIGINAL_CODE
 #define SOUND_SYSTEM_TYPE SoundSystem
+#else
+#include "..\..\platforms\Platform_Definitions.hpp"
```
iProgramMC: Name it PlatformDefinitions.hpp

platforms/Platform_Definitions.hpp
```diff
@@ -0,0 +1,12 @@
+#include <SoundSystem.hpp>
+
+//				Sound Systems
+//-----------------------------------------
+
+// Windows Sound System (Direct Sound)
+#if 1
```
iProgramMC: Use platform specific defines, for example, `#ifdef _WIN32`, `#elif defined(USE_SDL)` etc.

Vruk11: I think the issue with using platform specific platforms like _WIN32 or otherwise is that you might not want to use a platform specific code and use one that is cross-platform, for example I am making an OpenAL implementation and I might want to use that over the windows specific sound system or vice versa

iProgramMC: Okay, sure, have the OpenAL take over if USE_OPENAL is enabled for instance

Vruk11:
Sorry for all the force pushes, forgot to remove some stuff referencing OpenAL. Let me know if there's anything else that needs to be changed before this can be pulled @iProgramMC .

iProgramMC:
Can't merge as is, but I will make your changes in the main repo.

PR #20, opened 2023-08-02
PCM Sound File Loading

Kleadron:
I added a way for the game to load external .PCM sound sample files extracted from the libminecraftpe.so executable.
The scripts are now `grabsounds_header.py` and `grabsounds_pcm.py` and the PCM version extracts to the `sound_data_pcm/` directory.

The functionality can be disabled by commenting out `#USE_PCM_FILES` in `source/Sounds/SoundData.hpp`. I also updated the README.md to reflect this. Functionality of the PCM loader is also disabled when using the `ORIGINAL_CODE` define.

You need to copy the `.pcm` files from `sound_data_pcm/` to `assets/sounds/` in the game execution folder to use the feature, `SoundEngine`'s constructor calls `AppPlaform::loadSound` and needs to be created in `Minecraft::init` instead of the constructor since `m_pAppPlatform` isn't set when the constructor is called.
If the files cannot be found, the game loads empty sound data and will simply not play anything. I think this is more ideal than not being able to compile the game at all, and also means someone can easily make a .pcm file converter to add their own custom sounds.

`AppPlatform` has the new `loadSound` function that takes the path of the pcm file you want to use. I guess other codecs could be added through this but the file extensions would need to be reflected through `SoundEngine`. An implementation was created for `AppPlatform_windows` already.


iProgramMC:
Honestly, I'll consider pulling this after the split between the main modded version and the original decomp.

iProgramMC:
Please rebase to the latest commits

Kleadron:
I cleaned up the code with my latest version of it and ensured it matches with the master branch.

iProgramMC:
Hmm, wouldn't it be nicer to load actual wav files? Not as difficult as it seems as they're actually just raw PCM data with a header. [Here's an implementation of a very simple wav file reader.](https://github.com/iProgramMC/NanoShellOS/blob/master/apps/WavePlayer/src/main.c)

Kleadron:
Since this pull request seems to be inadequate and I currently don't feel like fixing it every time it has a new conflict I'm just going to close it.

PR #22, opened 2023-08-04
Names for Online

XDOneDude:
so basically the network.ini saves the player username and you can modify the ini to make a custom username, also its me flash


iProgramMC:
Sorry, but probably won't merge this.

I've made a commit which makes it so the user name mirrors the one in the options.txt.

PR #23, opened 2023-08-05
SDL/WASM Port

TheBrokenRail:
Sorry for the big PR! This PR adds an SDL port that can be used to build ReMinecraftPE for both Linux and WASM. The SDL port could probably also be used for Windows and Android with some minor modifications as well.

Summary of changes:
- `platforms/sdl`: The new SDL port. This is based on a combination of the existing Windows port and I stole the OpenAL code from my MCPI project.
- `build-wasm.sh`: Script for building the WASM port (only tested on Linux).
   - This uses an older version of Emscripten (does anyone remember how to spell that without copy/pasting it?) because of emscripten-core/emscripten#19921, when the fix is released in a stable version, the script should be changed to use `latest`.
- `compat/AKeyCodes.hpp`: Added a set of keycodes for SDL.
   - The Windows backend just makes the MCPE keycodes equal to Windows keycodes, but this doesn't work for SDL because some SDL keycodes are higher than 255, which breaks MCPE. Instead, the port has a set of MCPE keycodes and a function that converts a given SDL keycode to its equivalent MCPE keycode.
- `compat/GL.hpp`: This file has been modified to allow linking to Linux GL libraries.
   - There are two ways you can compile the SDL port:
      - OpenGL Desktop 2.1 Mode: Pretty self-explanatory, and similar to the Windows port.
      - Using the [OpenGL ES Compatibility Layer](https://gitea.thebrokenrail.com/minecraft-pi-reborn/gles-compatibility-layer): This bit of code stolen from my MCPI project emulates OpenGL ES 1.1 on top of OpenGL ES 3, which allows ReMinecraftPE to run on systems that don't support fixed-function OpenGL like WASM.
- `App/Minecraft.cpp`:
   - Stopped it from trying to access the network when building for WASM.
   - Blocked `setScreen(nullptr)` when a level doesn't exist, as otherwise it just locks up the game.
   - Some areas directly used Windows keycodes, those have been changed to use the correct keycode constant (ie. `AKEYCODE_1`).
- `Base/Utils.cpp`: Use the system's `opendir`/`readdir` when building for SDL, rather than a custom implementation.
- `GUI/GuiComponent.hpp`: Initialize `field_4` to `0`. This variable seems to be the Z co-ordinate for `GuiComponent`s, but it is *never* initialized. It's a coin-flip whether it's `0` and the widget shows up or if it's some insane number and it doesn't. Frankly, I'm surprised this worked reliably on Windows.
- `GUI/TextInputBox.cpp`: SDL has proper text input events, and so those are used instead of guessing based on keycodes.
- `Renderer/Textures.cpp`: RGB textures are loaded with a correct format of `GL_RGB` instead of `GL_RGBA`.
- `UserInput/Keyboard.cpp`: Bug fix to prevent crashing if a keycode is greater than 255.
- `World/Renderer/LevelRenderer.cpp`: You can't `++` a `bool`. My compiler threw an error, so I changed it to just set it to `true`.
- `thirdparty/coi-serviceworker`: Service worker which adds the proper headers for multi-threaded WASM: https://github.com/gzuidhof/coi-serviceworker.
   - I used an older version because the latest doesn't work reliably on Firefox, I need to file a bug report for that.
- `thirdparty/gles-compatibility-layer`: Aforementioned OpenGL ES Compatibility Layer.
- A lot of `CMakeLists.txt`: I used CMake as the build system.


iProgramMC:
Is OpenGL ES 3 not backwards compatible with 1.1? Does it not implement 1.1's APIs?

TheBrokenRail:
> Is OpenGL ES 3 not backwards compatible with 1.1? Does it not implement 1.1's APIs?

ES 2+ uses shaders and doesn't support fixed-function, while ES 1.1 doesn't have shaders but does have fixed-function. Desktop GL 2 has both. WASM uses WebGL which is a subset of ES 2/3.

> There is no "correct" format - MCPE internally uses RGBA. Force your image loader to put out RGBA textures.

The original code seems to support `GL_RGB` textures, it just only changes `internalFormat`. I just added code to also make it also change `format`:

```c++
	GLuint internalFormat = GL_RGB;

	if (texture.field_C)
		internalFormat = GL_RGBA;

	glTexImage2D(GL_TEXTURE_2D, 0, internalFormat, texture.m_width, texture.m_height, 0, GL_RGBA, GL_UNSIGNED_BYTE, texture.m_pixels);
```

> Please don't post a compiled build of the game with all the assets.

Sorry!

> Trying to use the scroll wheel crashes the game with the following stack trace.

Yeah, I noticed that too. Considering that it doesn't happen on a desktop (or at least happens silently), I assumed it was a bug with Emscripten or the existing scroll-wheel code. All the SDL code does is:

```c++
			case SDL_MOUSEWHEEL: {
				Mouse::feed(3, event.wheel.y, Mouse::_x, Mouse::_y);
				break;
			}
```

> It seems to only work if I get it from a locally hosted webserver. Is this intentional?

It should work on a non-local web server as long as it's HTTPS. Emscripten uses `SharedArrayBuffer` for multi-threading and that requires the page to be in a [secure context](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts).

platforms/sdl/AppPlatform_sdl.cpp
```diff
@@ -0,0 +1,308 @@
...
+	// Get Directory
+	std::string screenshots = _storageDir + "/screenshots";
+
+	// Get Timestamp
+	time_t rawtime;
+	struct tm *timeinfo;
+	time(&rawtime);
+	timeinfo = localtime(&rawtime);
+	char time[TIME_SIZE];
+	strftime(time, TIME_SIZE, "%Y-%m-%d_%H.%M.%S", timeinfo);
```
iProgramMC: Better to allocate plenty. We're not in the 1990s anymore where we had to watch for every little memory hog :)

compat/AKeyCodes.hpp
```diff
@@ -73,6 +74,109 @@ enum
...
+	AKEYCODE_A,
+	AKEYCODE_Q,
+	AKEYCODE_T,
+	AKEYCODE_Z,
+	AKEYCODE_F4,
+	AKEYCODE_ARROW_LEFT,
+	AKEYCODE_ARROW_RIGHT
+};
+
+static inline int translate_sdl_key_to_mcpe(int key) {
```
iProgramMC, line 131: Style does not match PE.

platforms/sdl/SoundSystemAL.cpp
```diff
@@ -0,0 +1,245 @@
...
+	{ \
+		ALenum __err = val; \
+		if (__err != AL_NO_ERROR) { \
+			LogMsg("(%s:%i) OpenAL Error: %s", __FILE__, __LINE__, alGetString(__err)); \
+			exit(EXIT_FAILURE); \
+		} \
+	}
+
+// Delete Sources
+void SoundSystemAL::delete_sources() {
```
iProgramMC: Style doesn't match again.

platforms/sdl/main.cpp
```diff
@@ -0,0 +1,260 @@
...
+			case SDL_QUIT: {
+				g_pApp->quit();
+				break;
+			}
+		}
+	}
+}
+
+// GUI Scale
+static void calulate_gui_scale() {
```
iProgramMC: Same thing. I probably will stop commenting on the style now

iProgramMC:
> It should work on a non-local web server as long as it's HTTPS.

No I meant that you can't load it by opening the html directly, it has to be hosted on a web server.

platforms/sdl/SoundSystemAL.cpp
```diff
@@ -0,0 +1,245 @@
...
+
+		// Store
+		buffers[sound.m_pData] = buffer;
+		return buffer;
+	}
+}
+
+void SoundSystemAL::update(float x, float y, float z, float yaw) {
+	// Check
+	if (loaded) {
```
iProgramMC: Use a guard clause instead. As in:
```cpp
if (!loaded)
    return;
```

TheBrokenRail:
> > It should work on a non-local web server as long as it's HTTPS.
> 
> No I meant that you can't load it by opening the html directly, it has to be hosted on a web server.

I'd assume that's just security stuff being weird with `file://`, it's not intentional, but I doubt there's anything either of us can do about it.

iProgramMC:
Also how does the TextInputBox::keyPressed call work? As far as you've told me you have separate code to translate sdl keycodes to "MCPE" ones. How does that translate in that function?

TheBrokenRail:
> Also how does the TextInputBox::keyPressed call work? As far as you've told me you have separate code to translate sdl keycodes to "MCPE" ones. How does that translate in that function?

SDL has a [specific API](https://wiki.libsdl.org/SDL2/SDL_TextInputEvent) for text events. So rather than trying to guess what character a keycode represents based on shift being pressed, it just uses that event to directly forward the characters to the text inputs:

```c++
			case SDL_TEXTINPUT: {
				if (g_pApp != nullptr && g_pApp->field_D14 != nullptr) {
					char x = event.text.text[0];
					if (x >= ' ' && x <= '~') {
						for (TextInputBox *input : g_pApp->field_D14->m_textInputs) {
							input->charPressed(x);
						}
					}
				}
				break;
			}
```

source/Renderer/Textures.cpp
```diff
@@ -69,7 +71,13 @@ int Textures::assignTexture(const std::string& name, Texture& texture)
 	if (texture.field_C)
 		internalFormat = GL_RGBA;
 
-	glTexImage2D(GL_TEXTURE_2D, 0, internalFormat, texture.m_width, texture.m_height, 0, internalFormat, GL_UNSIGNED_BYTE, texture.m_pixels);
+#ifdef ORIGINAL_CODE
+	GLuint format = GL_RGBA;
+#else
+	GLuint format = internalFormat;
```
iProgramMC: Don't do this, the dynamic texture patching code expects GL_RGBA, and there's no reason to 'fix' something that already works fine.

iProgramMC:
Also seems like networking just doesn't work. May want to disable the "Join Game" button on emscripten and remove the RakNet stuff if it just isn't supported?

TheBrokenRail:
> Also seems like networking just doesn't work. May want to disable the "Join Game" button on emscripten and remove the RakNet stuff if it just isn't supported?

Yeah, web stuff only has access to WebSockets and HTTP calls, while MCPE uses UDP sockets. I can disable the Join Game button, but I imagine untangling RakNet from MCPE would be quite a bit of effort for not much of a benefit.

iProgramMC:
Probably just need to disable everything that isn't required by RakNet BitStream.

iProgramMC:
One more thing, I should rename the `windows_vs` dir to `build` and then you can put your CMakeLists in there as well.

TheBrokenRail:
How would that be organized?

Right now, there's a fairly nice split, with a `CMakeLists.txt` for each "component":
- `source/CMakeLists.txt`
- `thirdparty/gles-compatibility-layer/CMakeLists.txt`
- `thirdparty/raknet/CMakeLists.txt`
- `platforms/sdl/CMakeLists.txt`

How would putting all of that in one folder work? Especially since keeping the components split-up as they are now helps keep the complexity down rather than having just one mega `CMakeLists.txt`.

iProgramMC:
Well, you know, there might be CMakeLists or Makefiles for different platforms...

iProgramMC:
Sorry man, I won't be able to pull this... Should have actually pulled break-core's changes at the end. Hope you don't mind, but please try with the new source code layout. This won't happen again.

platforms/sdl/AppPlatform_sdl.cpp
TheBrokenRail: Fair.

compat/AKeyCodes.hpp
TheBrokenRail: I adjusted all the braces to match the project's style.

platforms/sdl/SoundSystemAL.cpp
TheBrokenRail: Done.

source/Renderer/Textures.cpp
TheBrokenRail: Fixed.

compat/AKeyCodes.hpp
iProgramMC: I was mostly talking about the function name style, but I'll fix it myself.

iProgramMC:
LGTM!

PR #27, opened 2023-08-05
SoundSystemWindows Fix low volume on 3D Sounds, remove useless calloc. 

Stommm:
Mojang made the volume on 3D sounds very quiet so i multiplied it by 5 and it seems to match better to the 2D sounds now. The calloc crap was not needed i dont know what i was thinking


iProgramMC:
Looks good, thanks.

PR #29, opened 2023-08-05
Partial C++03 & Xbox 360 Support

BrentDaMage:
This took me a week to do. Please let me know if this may interfere with your workflow.


BrentDaMage:
Most of these changes consist of supporting the lack of foreach loops and delegating constructors.
I based my use of init and _init functions off of [this](https://stackoverflow.com/questions/308276/can-i-call-a-constructor-from-another-constructor-do-constructor-chaining-in-c) StackOverflow post.

iProgramMC:
What does 'partial' mean?

BrentDaMage:
As in it doesn't fully support C++03 just yet. The work I've done thus far is something that should be pushed to master so it doesn't have to be actively maintained alongside it.

PR #34, opened 2023-08-07
OpenAL Seperation and Adjustments

Vruk11:
Seperated OpenAL from SDL as much as possible as well ass removed some `#ifdef USE_SDL` statements that were not needed so that the OpenAL sound system works as a sound system without any extra code that was added by the SDL addition.
Also made some improvements to the code for the OpenAL sound system.
Also made PlatformDefinitions.hpp use "`USE_OPENAL`" instead of "`USE_SDL`" so that it works even if not being compiled for SDL.

The requirements for enabling OpenAL in the PlatformDefinitions.hpp is just to copy the include and libs folder from the OpenAL SDK into `thirdparty/OpenAL`.

The rest is dealt with by these commits adding: a `#pragma comment(lib, "OpenAL32.lib")` in the files that need the library, changing the file paths to the headers that are needed for the OpenAL sound system files to `thirdparty/OpenAl/include`, and adding an additional library path into the windows solution file pointing to `thirdparty/OpenAL/libs/Win64` and `thirdparty/OpenAL/libs/Win32` for their respective build architectures.

I am unsure if the seperation of OpenAL from SDL will break the CMake stuff for SDL but I attempted to avoid this by adjusting the path to the OpenAL folder. This might not be correct but I am not sure what I need to do for it and I haven't received a response yet.


platforms/openal/SoundSystemAL.cpp
```diff
@@ -226,73 +239,76 @@ void SoundSystemAL::update(float x, float y, float z, float yaw)
 	}
 }
 
-void SoundSystemAL::play(const SoundDesc& sound, float x, float y, float z, float volume, float pitch, bool is_ui)
+void SoundSystemAL::playAt(const SoundDesc& sound, float x, float y, float z, float volume, float pitch)
```
TheBrokenRail, line 243: Why did you remove `is_ui`? It's pretty important for UI sounds to work properly.

platforms/openal/SoundSystemAL.cpp
```diff
@@ -226,73 +239,76 @@ void SoundSystemAL::update(float x, float y, float z, float yaw)
...
-		alSourcef(al_source, AL_MAX_DISTANCE, 16.0f);
-		AL_ERROR_CHECK();
-		alSourcef(al_source, AL_ROLLOFF_FACTOR, 6.0f);
-		AL_ERROR_CHECK();
-		alSourcef(al_source, AL_REFERENCE_DISTANCE, 5.0f);
-		AL_ERROR_CHECK();
+	// Set Attenuation
+	alSourcef(al_source, AL_MAX_DISTANCE, 16.0f);
+	AL_ERROR_CHECK();
+	alSourcef(al_source, AL_ROLLOFF_FACTOR, 3.0f);
```
TheBrokenRail: Why did you change `AL_ROLLOFF_FACTOR`?

platforms/openal/SoundSystemAL.cpp
Vruk11: It doesn't actually affect anything in my experience, all UI sounds play perfectly that I have seen. It also requires extra code to be added related specifically to one sound system. that changes the SoundSystem interface that all SoundSystems should inherit off of.

platforms/openal/SoundSystemAL.hpp
```diff
@@ -0,0 +1,38 @@
...
+#include <thirdparty/OpenAL/Include/al.h>
+#include <thirdparty/OpenAL/Include/alc.h>
+
+#include <string>
+#include <vector>
+#include <unordered_map>
+
+#include "client/sound/SoundSystem.hpp"
+
+#pragma comment( lib, "OpenAl32.lib" )
```
TheBrokenRail: This should be in a `#ifdef _WIN32` block.

platforms/openal/SoundSystemAL.cpp
Vruk11: in https://github.com/ReMinecraftPE/mcpe/pull/34/commits/bc983340d61c9f9351289b0e12e19b4c10357e38
"Adjusted rolloff factor, not sure if this is just placebo but it felt a bit better?"

platforms/openal/SoundSystemAL.hpp
```diff
@@ -0,0 +1,38 @@
+#pragma once
+
+#include <thirdparty/OpenAL/Include/al.h>
+#include <thirdparty/OpenAL/Include/alc.h>
```
TheBrokenRail, line 5: There should be an `#ifdef _WIN32` block so it can use normal headers on Linux.

Vruk11: Normal headers? I don't understand.

platforms/openal/SoundSystemAL.cpp
TheBrokenRail: That really shouldn't work. UI sounds play at [0, 0, 0], so if a player is at [64, 64, 64] for instance, they won't hear UI sounds without disabling attenuation. Did you test if UI sounds work in-game?

Vruk11: Yes I did, it works just fine. I know it seems odd logically but it works just fine. Perhaps we could just have the function that plays the sounds pass the position of the player for the sound in an ifndef ORIGINAL_CODE block just in case?

platforms/openal/SoundSystemAL.hpp
TheBrokenRail: Like this:

```c
#ifdef _WIN32
#include <thirdparty/OpenAL/Include/al.h>
#include <thirdparty/OpenAL/Include/alc.h>
#else
#include <AL/al.h>
#include <AL/alc.h>
#endif
```

This way Linux users don't have to manually copy in header files.

platforms/openal/SoundSystemAL.hpp
Vruk11: Is there anything needed for other platforms to include the libraries?

platforms/openal/SoundSystemAL.cpp
TheBrokenRail: I built my own copy and have a theory. It seems stereo sounds (like `random.click`) have attenuation disabled automatically. So this works by pure chance. But if someone tried to play a mono sound in the UI, it wouldn't work. Or if someone had weird OpenAL drivers, it wouldn't work. So while this does work, personally I'd prefer the 4-5 extra lines to guarantee that it *will* work *always*, just for my own peace of mind. Just my opinion though.

platforms/openal/SoundSystemAL.hpp
TheBrokenRail: That's what [this piece of CMake](https://github.com/ReMinecraftPE/mcpe/blob/14562fc8cef55954681c6d1c591922ca81931fdf/source/CMakeLists.txt#L249-L255) does. On that note, `USE_OPENAL` has to be added to https://github.com/ReMinecraftPE/mcpe/blob/14562fc8cef55954681c6d1c591922ca81931fdf/platforms/sdl/CMakeLists.txt#L5.

platforms/openal/SoundSystemAL.cpp
Vruk11: I would prefer something for this as well but I think the only solution is to either have code that is specific to OpenAl (not really ideal) or to change the code to inherently support passing a boolean for UI sounds which would not be ideal due to the need to keep custom code in ORIGINAL_CODE blocks. And if we did use ORIGINAL_CODE blocks then we would have to implement each sound system to have a play function with and without taking in the UI boolean which is not ideal either

iProgramMC:
Looks good to me.

PR #36, opened 2023-08-07
Fixed rendering bugs & crashes

BrentDaMage:
* Fixed shading of cubes (includes Steve & Camera head)
* Fixed crash from rendering of Air item in hand
* Fixed crash from right-clicking with Air item in-hand
* TripodCameraRenderer no longer uses Cube::renderHorrible

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -385,3 +385,4 @@ FodyWeavers.xsd
 /assets/
 /options.txt
 /windows_vs/options.txt
+/games/com.mojang/minecraftWorlds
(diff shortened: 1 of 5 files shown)


BrentDaMage:
Screwed up the first commit. Apologies. I also added /games/ to the .gitignore for those using the root directory as the running directory for the game binary.

iProgramMC:
How do you even get "air" in your hand?

BrentDaMage:
I was able to get it in there via clicking one of the empty slots in the survival inventory. Even if that's unintended behavior and the survival inventory were fixed, this would still act as a failsafe.

iProgramMC:
Ah, okay. Anyways, it looks to me like you just turned off shading and made glColor4f work as intended instead. I'll just merge this and then fix the shading manually.

PR #37, opened 2023-08-07
C++03 Support & Partial Xbox 360 Support

BrentDaMage:
This should more or less support C++03. Most of these changes consist of supporting the lack of foreach loops and delegating constructors in C++03. I based my use of init and _init functions off of [this](https://stackoverflow.com/questions/308276/can-i-call-a-constructor-from-another-constructor-do-constructor-chaining-in-c) StackOverflow post.

This took me about a week to do. Please let me know if this may interfere with your workflow. It currently does not fully support the Xbox 360 due to the 360's lack of OpenGL compatibility.


BrentDaMage:
To be clear, none of the Xbox 360 support here should affect any other platform or any other workflows. Its presence relies on the presence _XBOX macro.

PR #39, opened 2023-08-08
Fix new entity ID system

EpixScripts:


iProgramMC:
Wow, what an oversight, thanks.

PR #41, opened 2023-08-08
Cleared up some fields

Vruk11:
Renamed some fields
Made some statements involving those fields make more sense for their type
Renamed the parameters of `SoundEngine::PlayAt` to have more readable names

### DIFF
--- a/source/Minecraft.cpp
+++ b/source/Minecraft.cpp
@@ -452,10 +452,7 @@ void Minecraft::tickInput()
 			}
 			else if (keyCode == AKEYCODE_SEARCH)
 			{
-				m_options.m_iPerspective++;
-				// If the perspective goes over the max loop back
-				if (m_options.m_iPerspective > 1) 
-					m_options.m_iPerspective = 0;
+				m_options.m_bThirdPerson = !m_options.m_bThirdPerson;
 			}
 			else if (keyCode == AKEYCODE_MENU)
 			{
--- a/source/Minecraft.cpp
+++ b/source/Minecraft.cpp
@@ -452,7 +452,10 @@ void Minecraft::tickInput()
 			}
 			else if (keyCode == AKEYCODE_SEARCH)
 			{
-				m_options.field_23D ^= 1;
+				m_options.m_iPerspective++;
+				// If the perspective goes over the max loop back
+				if (m_options.m_iPerspective > 1) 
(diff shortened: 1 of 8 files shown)


source/Minecraft.cpp
```diff
@@ -452,7 +452,10 @@ void Minecraft::tickInput()
 			}
 			else if (keyCode == AKEYCODE_SEARCH)
 			{
-				m_options.field_23D ^= 1;
+				m_options.m_iPerspective++;
+				// If the perspective goes over the max loop back
+				if (m_options.m_iPerspective > 1) 
+					m_options.m_iPerspective = 0;
```
iProgramMC: Please revert this back to the binary xor, or use `x = !x;`. You can't actually `++` a boolean. Also it's probably `m_bThirdPerson`.

While you're at it, change `field_23D`/`m_iPerspective` to a bool. It's actually a bool, not a uint8_t, but I didnt bother changing it back.

source/client/sound/SoundEngine.cpp
```diff
@@ -70,14 +70,14 @@ void SoundEngine::play(const std::string& name)
 	}
 }
 
-void SoundEngine::play(const std::string& name, float a, float b, float c, float d, float e)
+void SoundEngine::play(const std::string& name, float x, float y, float z, float volume, float pitch)
 {
-	if (m_pOptions->field_4 == 0.0f || d <= 0.0f)
+	if (m_pOptions->m_flMasterVolume == 0.0f || volume <= 0.0f)
```
iProgramMC: `m_fMasterVolume`.

source/Minecraft.cpp
Vruk11: I was thinking that it would be more ideal to have it as an int and have it as perspective in the inevitable case anyone wants to implement the front-facing thirdperson perspective. Do you still want me to have it be a boolean or is that a reasonable assumption and change?

iProgramMC: Yes, I still want it to be a boolean.

Vruk11: fixed in https://github.com/ReMinecraftPE/mcpe/pull/41/commits/fa385f886455a6f0023ca7abb39467506d20214f

iProgramMC:
Thanks!

PR #42, opened 2023-08-08
Fixed incorrect Entity push velocity direction

Marioiscool246:
When I was trying multiplayer out with two instances, I noticed that the push directions between the two players weren't correct. The Z axis wasn't being used for the push function that adds on the velocity. I've fixed this below while also making sure that the bug itself wasn't an actual bug in mcpe 0.1.0 by looking at the disassembly from the apk.


iProgramMC:
Thanks - even if it were a bug in the original version of the game we still fix such bugs. This is no longer accurate to the demo - see [ReMinecraftPE/mcped](https://github.com/ReMinecraftPE/mcped) for the accurate decomp version.

Marioiscool246:
This fix would also apply to the accurate decomp version.

PR #44, opened 2023-08-09
Fix incorrect axes being used for clip min/max

Marioiscool246:
I've noticed that sometimes the pick hit result would go through entities on certain axes except X. After some investigation and debugging, I've fixed the issue. It seems like this was caused by an error in decompilation, similar to my other issue (#42) from examining the code from the apk.


iProgramMC:
BTW, other issues have been fixed in the aabb/tile code. Now tile picking is perfect!!

Marioiscool246:
Great!

PR #45, opened 2023-08-09
Fix held item not always being changed to the new held item

Marioiscool246:
`pPlayer->m_pInventory->selectSlot(0);` overwrites the selected slot that is changed inside `pPlayer->m_pInventory->selectItemById`.


iProgramMC:
Nice, thanks for the fix :)

PR #46, opened 2023-08-09
Fix empty inventories for server players

Marioiscool246:
Have prepare inventory functions be called by the `Inventory` constructor instead of elsewhere. If survival mode is worked on further, this'll most likely have to be changed again.

### DIFF
--- a/source/client/network/ClientSideNetworkHandler.cpp
+++ b/source/client/network/ClientSideNetworkHandler.cpp
@@ -104,6 +104,8 @@ void ClientSideNetworkHandler::handle(const RakNet::RakNetGUID& rakGuid, StartGa
 		pLocalPlayer->m_yaw,
 		pLocalPlayer->m_pitch);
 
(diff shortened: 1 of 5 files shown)


iProgramMC:
Don't. It should be initialized, please figure out WHY it isn't instead of initializing it elsewhere.

Marioiscool246:
I thought it would've been a good idea to mimic the original game's behavior, but I was wrong. Sorry about that. This one should be 100 times better though.

Marioiscool246:
Also, what's the recommended bracket style? I'm seeing different kinds.

iProgramMC:
The preferred style is opening brace on new line, as in 
```cpp
void SomeFunction()
{
    ....
}
```

iProgramMC:
> I thought it would've been a good idea to mimic the original game's behavior, but I was wrong. Sorry about that. This one should be 100 times better though.

This is in the original?

PR #51, opened 2023-08-12
Fixed double destroy sounds/particles

Vruk11:
I decided to fix it in the creative gamemode specifically to reduce impact on other gamemodes but the `Minecraft::handleMouseClick` and `Minecraft::handleMouseDown` might still need a fix for `field_DA4` in the future, not sure.

This code might slow down the speed at which blocks are destroyed when holding down the mouse button but only because the game was trying to destroy blocks like.. twice per tick before?


iProgramMC:
I've made a breaking change - removed (almost) all the SurvivalMode-like code in CreativeMode and override just `destroyBlock`, which fixes this issue. Thanks for the PR, but it won't get merged.

Issue #59, opened 2023-08-17
Add Hide GUI option


JasiuWu2006:
Please add it either it can be hidden by f1 like in modern minecraft or as a toggleable option in the menu.

iProgramMC:
not an issue, add it yourself

BrentDaMage:
I think it'd be good to track what features should be added via issues. This one is pretty minor and having the issue here could help someone who's looking for something to do.

BrentDaMage:
Or perhaps do feature requests on the Discord in the form of threads so people can expand on their ideas? No promises would need to be made that anything gets done, it'd just act as some inspiration for people looking for something to work on.

iProgramMC:
Know what, sure.

PR #62, opened 2023-08-19
Added Hide GUI Keybind

Vruk11:
Adds feature requested by https://github.com/ReMinecraftPE/mcpe/issues/59
F1 to hide GUI
Also renamed the field for hiding the GUI to `m_bDontRenderGui`

I would have named it `m_bRenderGui` and flipped the default values and checks but I assumed that changing the least amount of stuff would be desired.


iProgramMC:
looks good to me :^)

Issue #60, opened 2023-08-17
Add keybinds.


JasiuWu2006:
As the title says. It would be cool if you could rebind the keys.
Sorry for abusing Issues tab.

iProgramMC:
Please add it yourself or wait for someone else to add it.

PR #66, opened 2023-08-23
Updated keymapping and added controls menu

Vruk11:
Added` SetDefaultBindings` to`AppPlatform`, Created a `AppPlatform_Windows` and `AppPlatform_sdlbase` implementation. This runs from `Minecraft::init()`
Added `ControlsScreen`, a scrolling menu that dynamically adds buttons for remapping keys 
Added a new controls button in the `OptionsScreen`
Removed struct `KeyBind` from class `Options`

Solves https://github.com/ReMinecraftPE/mcpe/issues/60


iProgramMC:
List of suggested changes (can just apply in git)

[suggestedchanges.patch](https://github.com/ReMinecraftPE/mcpe/files/12416151/suggestedchanges.patch)

Also:
* Make the 'controls' button centered and add ellipses to show that it requires further interaction
* Use two columns instead of one
* Add a new virtual to the AppPlatform which will translate a key code into a user friendly string
* Remove some of these, or move them to the bottom. Beta 1.3 only allowed you to change some of the keybinds. For reference, here's a list of custom keybinds supported in Beta 1.3
image

source/client/gui/screens/ControlsScreen.cpp
```diff
@@ -0,0 +1,245 @@
...
+}
+
+void ControlsScreen::render(int a, int b, float c)
+{
+	if (!m_pMinecraft->isLevelGenerated())
+		renderMenuBackground(c);
+
+	fillGradient(0, TOP_CUTOFF - 2, m_width, m_height - BOTTOM_CUTOFF + 7, 0xC0101010, 0xD0101010);
+	//Won't work?
+	//fill(0, 0, m_width, TOP_CUTOFF, 0xFFFFFF);
```
uniformization, line 190: Don't know if this was intentional because of [here](https://discord.com/channels/1135450987096703096/1135450987549704234/1143715947870433300) but I would recommend using `renderBackground()` when we're in a level.
```cpp
if (!m_pMinecraft->isLevelGenerated())
{
	renderMenuBackground(c);
	fillGradient(0, TOP_CUTOFF - 2, m_width, m_height - BOTTOM_CUTOFF + 7, 0xC0101010, 0xD0101010);
}
else
{
	renderBackground();
}
```

Result:
UIhcgFAfOk

uniformization:
Keybinds do not save when I change them or close the game.

uniformization:
This is just my opinion but you could copy/modify `ScrolledSelectionList` to go for the background effect I think you tried to go for.
bO1UT2nKJH
mYNu6dpOio

iProgramMC:
> This is just my opinion but you could copy/modify `ScrolledSelectionList` to go for the background effect I think you tried to go for. 

You can subclass it. See an example in AvailableGamesList`.

iProgramMC:
Hey man, I won't merge this PR as it's working on an extremely outdated source code base. You can submit a new PR with similar changes.

PR #68, opened 2023-08-26
Mac OS X 10.6 & More C++03 Support

BrentDaMage:
We didn't have proper C++03 support previously.
Screen shot 2023-08-25 at 12 26 01 PM


iProgramMC:
> `// TODO: This isn't specific to SDL2. Why isn't it in an AppPlatform base class?`

Because on android for instance, the options are stored somewhere else. But thanks

BrentDaMage:
> > `// TODO: This isn't specific to SDL2. Why isn't it in an AppPlatform base class?`
> 
> Because on android for instance, the options are stored somewhere else. But thanks

That doesn't mean you need to copy-paste the code for loading said file across multiple classes. You can just pass the options.txt path into the Options class for example and have Options take it from there.

PR #69, opened 2023-08-27
Logging cleanup

BrentDaMage:
* Added StandardOut class
* Renamed LOGX macros to LOG_X
* Removed LogMsg macros in favor of LOG_X
* Added console window for debug Windows builds


iProgramMC:
FYI the LOGX macros are aliased to printf/puts and then to LogMsg.

BrentDaMage:
I know how it works. I just went it and changed all of it.

iProgramMC:
I suggest you instead change it all to LogMsg.

BrentDaMage:
Can you please take a look at the changes I made before suggesting things? Thanks.

BrentDaMage:
Also, I have tested my changes on Windows 10 and Mac OS X 10.6. They both build and run as intended.

iProgramMC:
> Can you please take a look at the changes I made before suggesting things? Thanks.

Yes.



> Also, I have tested my changes on Windows 10 and Mac OS X 10.6. They both build and run as intended.

This is not the problem at hand. I want you to instead of changing them to LOG_X, change them all to LogMsg.

BrentDaMage:
No? LOG_X is better because you have the ability to specify what kind of message it is; info, warning, and error. LOG_I, LOG_W, LOG_E make infinitely more sense as macros because they are clearly identifiable as such. They are not able to be confused with functions. LogMsg is however.

iProgramMC:
LogMsg is indeed a function.

Also, it seems like you made the debug console show up as an actual window. I like it just the way it is, logging stuff to the _Visual Studio Debugger_ when running from within VS (through the OutputDebugString call). If you want you can make LogMsg output to the screen and **change the subsystem of the application to CONSOLE rather than WINDOWS**.

BrentDaMage:
LogMsg isn't a function anymore. It has been replaced with another function that's called via LOG_X. LOG_X is a macro which calls a function of an instantiated class. There is no longer a function for logging that resides in the global namespace (other than the standard C++ ones). The code for displaying the console is based on whether or not the project is compiled as debug. If it's compiled as debug, the console window will display. If this behavior can be replicated using the CONSOLE subsystem and you would like to see it implemented as such, you're welcome to implement it.

BrentDaMage:
Unless you can present to me a set of valid reasons as to why I should remove my LOG_X changes and revert them to LogMsg, I will not be doing so. I do not consider "I own the repo" to be a valid reason for me to refactor/undo my hours worth of work.

BrentDaMage:
So far it does not sound like you've actually reviewed my modifications to the codebase. I implore you to actually take a look at what functions I added/removed/changed regarding output before attempting to argue against my changes.

uniformization:
> LogMsg is indeed a function.

> I suggest you instead change it all to LogMsg.

Do you understand this PR? It’s to add logging levels. Minecraft Java does it too and it makes it easier to quickly skim through log files, especially if you are reading huge log files.

iProgramMC:
> It’s to add logging levels. Minecraft Java does it too and it makes it easier to quickly skim through log files, especially if you are reading huge log files.

Fine, but you can add helper functions such as LogError and LogWarning that do the same thing.

iProgramMC:
Also, there is absolutely no need to refactor the windows logging code. If you REALLY want a console window, instead of calling `AllocConsole()`, for debug builds, change the subsystem from Windows to Console, and then change the calls from OutputDebugStringA to puts.

iProgramMC:
The reason I overloaded LogMsg() in the first place (instead of defining it like `#define LogMsg(str, ...) printf(str "\n", __VA_ARGS__);` (which is fine), is because I specifically wanted windows debug output to be piped into the visual studio debug console (see image). It's very useful for debugging as the window that I'd normally need to also drag around with me is conveniently part of VS itself and thus I can dock it.

image

iProgramMC:
Also I recommend rebasing because I've merged #68.

BrentDaMage:
> > It’s to add logging levels. Minecraft Java does it too and it makes it easier to quickly skim through log files, especially if you are reading huge log files.
> 
> Fine, but one can add helper functions such as LogError and LogWarning that do the same thing.

Please for god's sake please look at what I changed. "LogError" and "LogWarning" essentially already exist in the form of LOG_E and LOG_W. I already added exactly what you described in the form of macros.

> Also, there is absolutely no need to refactor the windows logging code.
All printing is still going out through OutputDebugStringA() as well as stdout. The logging code functions fundamentally the same.

> If you REALLY want a console window, instead of calling AllocConsole(), for debug builds, change the subsystem from Windows to Console, and then change the calls from OutputDebugStringA() to puts().
Again, I would like to have the console window only appear when running as debug. Also, there is no reason to change any calls made to OutputDebugStringA(), because again, everything is printed to there and stdout as it is. You would know this if you actually read through my changes to the codebase.

> The reason I overloaded LogMsg() in the first place... is because I specifically wanted windows debug output to be piped into the visual studio debug console...
Ok cool. I like that feature too, that's why I kept it... If you had read the code you would realize that all logging is disabled if the client is not built as debug. Do you have a quarrel with that change? If so, please let me know. But just to reiterate, output still goes to the VS debug console. You would know that I kept it if you actually read the code.

iProgramMC:
> Please for god's sake please look at what I changed. "LogError" and "LogWarning" essentially already exist in the form of LOG_E and LOG_W. I already added exactly what you described in the form of macros.
Yep, exactly, but I don't want them as macros, I want them as functions.

> Ok cool. I like that feature too, that's why I kept it...
My bad.


Also a singleton for logging is probably useless. I'd just keep them as global functions.

BrentDaMage:
> Yep, exactly, but I don't want them as macros, I want them as functions.
Okay. Why? You realize the macros just call on functions? Having macros here will allow you to change what's called without having to rely on a specific function. For example, we're currently just prefixing all the LOG_W prints with "[WARNING]: " and calling StandardOut::singleton()->printf() but in the future we could call a completely different printf function with an enum that specifies the print to be a warning. This way we are not tied to a function in the global namespace.

> Also a singleton for logging is probably useless. I'd just keep them as global functions.
There's no reason to keep them as global functions. That is less organized. The point of the macros is to allow easy calling of the StandardOut singleton's functions. So LOG_X calls a function from an instantiated class and can be redirected anywhere, while LogWarning would always be function that exists in the global namespace under no class.

iProgramMC:
> [...] we're currently just prefixing all the LOG_W prints with "[WARNING]: " and calling StandardOut::singleton()->printf() but in the future we could call a completely different printf function with an enum that specifies the print to be a warning.

Why aren't you doing this, like, right now?

BrentDaMage:
> Why aren't you doing this, like, right now?
We are. That is why I said "we're currently". It's because that's how we're doing it in my changes. And yes, I agree that it would be cleaner than what we used to be doing. That's why I'm doing it.

BrentDaMage:
Additionally I've just added another commit that showcases the benefits of using objects rather than sticking functions in the global namespace. I'm able to override the default StandardOut class with a Windows-specific one that also prints things to the VS debug console (as it's been doing). So there's no longer a need to do the #ifdef check I had been doing in StandardOut.

iProgramMC:
Sure. I'll do the enum change then

Cristiano100:
@iProgramMC bro u good?

iProgramMC:
huh?

PR #71, opened 2023-08-27
Options Logic Cleanup

BrentDaMage:
* options.txt now loads from Options class.
* Moved options loading code from AppPlatform classes to Options class
* Added AppPlatform::singleton()
* Minecraft::m_options is now only accessible via Minecraft::getOptions() (as it should be)
* Options won't try to load if hasFileSystemAccess returns false. Emscripten build will be happy.


BrentDaMage:
Note to self to go in and fix all of the `..\..\..\`'s in the vcxproj files with ` $(MC_ROOT)`. These popped up because of the new files that were added to the project.

iProgramMC:
Please don't submit PRs on the same branch.  Now you've included 2 other PRs' worth of changes (#68 and #69).

BrentDaMage:
Then we can cross that bridge when we come to it. The way I see it, there is currently only one way to load options. Having the code for doing so in one place is much better than having it copy-pasted across 4 separate places. Mojang seems to agree on this and it can be seen in 0.13.0. If we really wanted to, we can expand on what I've done and add something like an `OptionsSource` class or some sort of generic class for loading properties from a file or source (like Mojang did). While what I've done may make things slightly more difficult if we decide to load options from a different source, my changes also ensure that a proper clean solution will be need to be implemented to solve said problem.

iProgramMC:
Actually not a bad idea. After a review I think this is fine.

BrentDaMage:
The branch is much less fucked up now in terms of its divergence from master. Not sure how to get rid of e35ee0d879ab3b4c964ec601c62cb1f0df45657c. Let me know if it's a problem. If so, any advice would be greatly appreciated.

PR #79, opened 2023-09-23
WIP Android Port

Stommm:
Work in progress Android port. 
Still needs touch controls and mouse turning (if that's even possible on android) and file saving and SoundSystemSL. 
You control the camera and movement with your controller for now. You can navigate the gui using touch. Options.cpp,LocalPlayer.cpp,Minecraft.cpp is configured to use controller on android. 
Blocked out some code in ControllerTurnInput.cpp,Controller.cpp that didn't make sense for android.
Fix clang compiler error in Mth.cpp. 
Using zlib that ships with android.
Create visual studio project and implement AppPlatform_android


source/client/renderer/GameRenderer.cpp
```diff
@@ -622,7 +626,9 @@ void GameRenderer::render(float f)
 
 	if (m_pMinecraft->m_pScreen)
 	{
-		glClear(GL_ACCUM);
+		#ifndef __ANDROID__
```
iProgramMC: Was a thing on the Android build. Why did you remove this?

source/client/renderer/GameRenderer.cpp
```diff
@@ -204,7 +204,9 @@ void GameRenderer::setupGuiScreen()
 {
 	float x = Gui::InvGuiScale * Minecraft::width;
 	float y = Gui::InvGuiScale * Minecraft::height;
+#ifndef __ANDROID__
 	glClear(GL_ACCUM);
```
iProgramMC: Was a thing on Android. Why remove this?

source/client/player/input/ControllerTurnInput.cpp
```diff
@@ -22,6 +22,9 @@ ControllerTurnInput::ControllerTurnInput()
 
 ITurnInput::Delta ControllerTurnInput::getTurnDelta()
 {
+#ifdef __ANDROID__
+	return Delta(Controller::getX(m_stickNo) * 50.f, Controller::getY(m_stickNo) * 60.f);
```
iProgramMC: Why 50 and 60? Magic values much?

Note: this code was written for the xperia play, so you will want to modify it, but I'm just checking that it's modified correctly

iProgramMC (commented):
Overall, looks good, though honestly, I'll probably pull this into a separate branch and work on it from there, and then it'll go to master.

TheBrokenRail:
Why not use SDL's Android support instead of writing a whole new backend?

iProgramMC:
Honestly I'd do the same thing. To be close to the original source layout I guess

source/client/renderer/GameRenderer.cpp
Stommm: it didn't really made much sense since GL_ACCUM does not exist in embedded opengl so i blocked it out. TheBrokenRail mentioned it uses GL_DEPTH_BUFFER_BIT

source/client/player/input/ControllerTurnInput.cpp
Stommm: Yeah.... its just some values to increase turn sensitivity it should probably be implement in Options

iProgramMC:
That would be reason 1. Reason 2 is what I stated above.

TheBrokenRail:
SDL's minimum Android version is API 19 (v4.4). This means it suppirts [99.64%](https://apilevels.com/) of devices.

BrentDaMage:
> SDL's minimum Android version is API 19 (v4.4). This means it suppirts [99.64%](https://apilevels.com/) of devices.

In that case I think SDL2 would be best suited for use on desktop Android machines (depending on how we implement the mobile-friendly UI). Usually we can get the most out of a platform by using its native APIs, so if someone's already attempting this, I don't see a reason not to go along with it. Additionally, it seems as though SDL 2.0.8 or below actually supports API 12 which is from around early 2011. Unfortunately iOS isn't at all similar. The oldest version SDL2 seems to support is iOS 8, and maaaybe iOS 6.1 at best.

iProgramMC:
I support creating a separate project because it's more faithful to the original code base, first and foremost. The rest are just details.

SanyaSho:
I remember when someone compiled source engine for android with NDK r10e and it didn't run on android 14. But when he updated NDK to r20* it ran without any problem.

TheBrokenRail:
My opinion is still that faithfully recreating every `AppPlatform` for every target just isn't worth it. Especially, since each `AppPlatform` also tends to require its own build-system.

I'm a little biased (since I wrote the SDL backend), but I think it would be way nicer to just use SDL when possible instead of having dedicated Windows/Android/iOS(?)/Linux backends. Less code is usually better in my opinion.

BrentDaMage:
We get more control over the platform if we write platform-specific AppPlatform classes. Doing one large AppPlatform_sdl.cpp file with a shitload of #ifdefs is not a viable solution to wanting more control over the platform either.

TheBrokenRail:
How much platform-specific control does MCPE even need?

Anyways, just like SDL does currently, it doesn't have to be just one SDL `AppPlatform`, there can be one base-class with multiple sub-classes. Also, unlike WASM, which is just kind of weird, Android/Linux/Windows are all fairly similar with regards to how MCPI interacts with them. If I'm not wrong, you've actually already ported the SDL backend to Windows and you can barely tell.

BrentDaMage:
Correct. I've already gotten the SDL2 app running under Windows. Only problem is, managing libraries (such as SDL2) under Windows is horrible. So we prefer to keep the option around to use built-in libraries.

iProgramMC:
Good news, I've added in the touch screen support. Please rebase and integrate my changes.

PR #80, opened 2023-09-23
Visual Studio Project Overhaul + Cleanup

BrentDaMage:
- SDL2 project for Windows
- Re-added game client icon to SDL2 code
- Renamed "AppPlatform_windows" to "AppPlatform_win32" (this is the name of the Windows API and is not representative of the architecture type)
- Renamed "LoggerWindows" to "LoggerWin32"
- Renamed "SoundSystemWindows to "SoundSystemDS" (DirectSound). This may be used for the 360, so it wouldn't really be Windows-specific then.
- Moved "ClientSideNetworkHandler" from "network" to "client/network". We don't need it being compiled for the server if the client's the only thing that needs it.
- Bugfixes for macOS
- Options::savePropertiesToFile Logging Bugfix
- Silence Winsock Deprecation Warnings in RakNet


BrentDaMage:
Planning on separating the projects and libraries and such out with XCode later. I'm not familiar with CMake at all, so if someone wants to make each folder into its own library to match the VS projects, that's fine by me.

iProgramMC:
The last complaint I have is that I can't seem to put breakpoints in the 'Tile' part of the 'World' project (I tried it in LiquidTileDynamic and it just won't pick it up!)

BrentDaMage:
> The last complaint I have is that I can't seem to put breakpoints in the 'Tile' part of the 'World' project (I tried it in LiquidTileDynamic and it just won't pick it up!)

Are you compiling it as Debug?

iProgramMC:
I wasn't. I addressed it on the Discord, sorry for not retracting my statement on here as well.

iProgramMC:
Note: We desperately need a place to download assets from. The v0.1.0 block textures are few and most are used. I'm considering adding a web based downloader which downloads assets from minecraft.net's official CDN.

iProgramMC:
Also note: Water is really heavy lately for some reason. Desperately needs a fix (from me) as well.

iProgramMC:
So aside from debug (which was like this before the pull), it works perfectly! Thank you!

PR #83, opened 2023-10-10
Allow Using Block Outline Instead Of Highlight

### LINKED ISSUE #72: Outline blocks instead of brightening them
It would be pretty cool if there was an option to outline selected blocks instead of making them brighter.
Please consider adding this.

TheBrokenRail:
Fixes #72

### DIFF
--- a/source/client/renderer/LevelRenderer.cpp
+++ b/source/client/renderer/LevelRenderer.cpp
@@ -1044,10 +1044,15 @@ void LevelRenderer::renderHitOutline(Player* pPlayer, const HitResult& hr, int i
 
 	glEnable(GL_BLEND);
 	glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);
-	glColor4f(1.0f, 1.0f, 1.0f, 0.4f);
-	glLineWidth(10.0f);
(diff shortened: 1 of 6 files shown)


TheBrokenRail:
Rebased onto `master`.

source/client/renderer/LevelRenderer.cpp
```diff
@@ -1044,11 +1044,16 @@ void LevelRenderer::renderHitOutline(Player* pPlayer, const HitResult& hr, int i
...
-	glLineWidth(10.0f);
+	glColor4f(0, 0, 0, 0.4f);
 	glDisable(GL_TEXTURE_2D);
 	glDepthMask(false);
 
+	// Maximize Line Width
+	glEnable(GL_LINE_SMOOTH);
+	int range[2];
+	glGetIntegerv(GL_SMOOTH_LINE_WIDTH_RANGE, range);
+	glLineWidth(range[1]);
```
iProgramMC, line 1055: I probably won't pull this directly, but I will integrate your changes. Why? It seems to me that things just break, I can't break any blocks.

Also this gets the maximum line width, which is 63 on Windows. Really we should be using a line width of 1, or a line width of 2 checked against the max line width range.

PR #85, opened 2023-10-11
Make ZLib A Submodule (And Use The Host's ZLib If Possible)

TheBrokenRail:
This makes ZLib a sub-module rather than just directly including it in the repo. This should make ZLib easier to update in the future.

It also updates the vendored ZLib from v1.2.13 to v1.3.

Also, if the system already has ZLib installed, CMake just use that.

The CMake project also now uses the [official ZLib CMake](https://github.com/madler/zlib/blob/develop/CMakeLists.txt), rather than a custom one.

This *might* break the VS project if ZLib v1.3 has a different source layout than v1.2.13. I don't have Windows so I can't check.


TheBrokenRail:
Rebased on `master`. This still almost certainly breaks the VS build.

PR #91, opened 2023-11-02
SDL Multitouch!

TheBrokenRail:
This adds multitouch support to the SDL build.

### DIFF
--- a/platforms/sdl/main.cpp
+++ b/platforms/sdl/main.cpp
@@ -14,6 +14,7 @@ typedef AppPlatform_sdl UsedAppPlatform;
 #endif
 
 #include "client/app/NinecraftApp.hpp"
+#include "client/player/input/Multitouch.hpp"
 
 #ifdef __EMSCRIPTEN__
 #include <emscripten.h>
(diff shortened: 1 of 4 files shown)


platforms/sdl/main.cpp
```diff
@@ -54,6 +55,60 @@ static int TranslateSDLKeyCodeToVirtual(int sdlCode)
 	return SDLVK_UNKNOWN;
 }
 
+// Touch
+#define TOUCH_IDS_SIZE (MAX_TOUCHES - 1) // ID 0 Is Reserved For The Mouse
```
iProgramMC, line 59: `Multitouch` does not reserve IDs for the mouse. Also the `Mouse` and the `Multitouch` are checked separately, have separate event queues, and are basically entirely separately, except for the fact that Multitouch is implemented on top of an array of `MouseDevice`s.  So I don't think you need this.

source/CMakeLists.txt
```diff
@@ -90,22 +90,22 @@ add_library(reminecraftpe-core STATIC
...
-	client/player/input/CustomInputHolder.cpp
-	client/player/input/IInputHolder.cpp
-	client/player/input/IMoveInput.cpp
-	client/player/input/ITouchScreenModel.cpp
-	client/player/input/TouchAreaModel.cpp
-	client/player/input/TouchInputHolder.cpp
-	client/player/input/TouchscreenInput_TestFps.cpp
-	client/player/input/UnifiedTurnBuild.cpp
     client/network/ClientSideNetworkHandler.cpp
+    client/player/input/IBuildInput.cpp
```
iProgramMC, line 93: Why were these moved to insert `ClientSideNetworkHandler.cpp` in the middle of the `client/player/input` group?

iProgramMC:
Other than these, and the fact that there's more changes than necessary due to you changing the indent characters, your PR looks fine to me.

source/CMakeLists.txt
TheBrokenRail: Because I started this PR before you fixed the CMake and the rebase messed things up. I'll fix it.

platforms/sdl/main.cpp
TheBrokenRail: If mouse events don't also send multitouch events, things become fairly buggy. But for mouse event callbacks to send multitouch events, they need a touch ID. So I reserved ID 0.

TheBrokenRail:
Sorry about the indents, I accidentally used 4-space indents when starting this PR, and fixing that seems to have fixed other code as well.

source/CMakeLists.txt
TheBrokenRail: This is fixed now.

PR #92, opened 2023-11-02
Android Port Using SDL

TheBrokenRail:
Personally, I don't like reinventing the wheel for every platform. As such, I'm not the biggest fan of how the existing WIP Andorid port does so, especially when SDL has perfectly good Android support. I especially don't like how that port depends on VS (and therefore Windows) to build.

So here's my Android port using the existing SDL backend. It supports both multi-touch and showing the on-screen keyboard when appropriate.

image


iProgramMC:
The currently WIP port will not require Windows to build when it's merged in. I will personally port it to a gradle based project. It's only to get us up and running.

iProgramMC:
Sure. I can pull your changes too as an alternative. I mean they build on your existing work anyway. I'd refuse if it was purely Android that's supported by SDL.

uniformization:
Best not to use Minecraft assets in `platforms/android/app/src/main/res`

iProgramMC:
> Best not to use Minecraft assets in `platforms/android/app/src/main/res`

Think there's already a low resolution graphic of the Minecraft grass block in Brent's Apple stuff. I should think of an icon to replace it.

iProgramMC:
Please remove the Minecraft: Pocket Edition icon. We don't want to risk it. It's fine if it's iconless for now - the native Android port has no icon.

Also, it seems that immersive mode is turned off when the virtual keyboard is shown.

Also, the virtual keyboard has text prediction enabled for some reason, which is broken.

TheBrokenRail:
I've made the changes!

PR #94, opened 2023-11-03
Replace ZLib With A Submodules (v3.0)

TheBrokenRail:
This PR does two things:
1. It modifies the CMake build to use the system's ZLib, rather than the vendored version.
2. It replaces the vendored version with a submodule of the same version.

One thing to note: the submodule version is ZLib is almost identical to the old version, except for one change:
```patch
--- zlib/gzguts.h
+++ zlib-old/gzguts.h
@@ -37,13 +37,15 @@
 
 #if defined(__TURBOC__) || defined(_MSC_VER) || defined(_WIN32)
 #  include <io.h>
+#else
+#  include <unistd.h>
 #endif
 
 #if defined(_WIN32)
 #  define WIDECHAR
 #endif
 
-#ifdef WINAPI_FAMILY
+#if defined(WINAPI_FAMILY) || defined(_WIN32)
 #  define open _open
 #  define read _read
 #  define write _write
```
I do not know why the old version of ZLib had undocumented changes, but the submodule version doesn't.


iProgramMC:
So how did you embed those changes into the submodule zlib?

TheBrokenRail:
I'm going to mark this as a draft for now

PR #95, opened 2023-11-03
Update Linux Build Instructions

TheBrokenRail:
This updates the build instructions for Linux to use the CMake build-system. It also, at least in my opinion, makes them clearer.


iProgramMC:
Cool

PR #96, opened 2023-11-03
Fix MCPE Not Covering Screen Edges

TheBrokenRail:
This fixes an issue where MCPE doesn't cover all of the screen's edges.

| Before | After |
| --- | --- |
| Screenshot from 2023-11-03 00-06-52 | Screenshot from 2023-11-03 00-09-07 |

### DIFF
--- a/source/client/renderer/GameRenderer.cpp
+++ b/source/client/renderer/GameRenderer.cpp
@@ -205,12 +205,12 @@ void GameRenderer::saveMatrices()
 
 void GameRenderer::setupGuiScreen()
 {
-	float x = Gui::InvGuiScale * Minecraft::width;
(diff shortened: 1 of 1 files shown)


iProgramMC:
After your change, the GUI is no longer pixel perfect, as it should, on screen resolutions that aren't a multiple of the GUI scale. While this is a non-issue, neither are the uncleared bottom right screen edges.

TheBrokenRail:
Later version of MCPE, have the same behavior as this PR:

Screenshot from 2023-11-03 10-12-09

Also, personally, the dirty screen edges bug me far more than the non pixel-perfect GUI that I don't even notice.

BrentDaMage:
I have a similar problem on older versions of Mac OS X using SDL2, except it's usually on the left and right side of the screen.

PR #97, opened 2023-11-03
HaikuOS port

SanyaSho:


thirdparty/raknet/SocketLayer.cpp
```diff
@@ -24,6 +24,10 @@
 #include <netdb.h>
 #endif
 
+#if defined( __HAIKU__ )
```
TheBrokenRail, line 27:
```suggestion
#ifdef __HAIKU__
```

platforms/sdl/CMakeLists.txt
```diff
@@ -20,6 +20,11 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
     add_compile_options(-Wno-inconsistent-missing-override -Wno-enum-compare-switch -Wno-register)
 endif()
 
+# Network library
+if(HAIKU)
+    add_link_options(-lnetwork)
```
TheBrokenRail:
```suggestion
    link_libraries(network)
```

source/world/level/storage/ExternalFileLevelStorageSource.cpp
```diff
@@ -58,7 +58,16 @@ void ExternalFileLevelStorageSource::getLevelList(std::vector<LevelSummary>& vls
 
 		LOG_I("Entry: %s", de->d_name);
 
+#if defined( __HAIKU__ )
+		// @HACK: make a temp buffer here
+		char tempbuf[PATH_MAX];
+		snprintf( tempbuf, sizeof( tempbuf ), "%s/%s", m_worldsPath.c_str(), de->d_name );
+
+		struct stat buf;
+		if ( ( lstat( tempbuf, &buf ) == 0 ) && S_ISDIR( buf.st_mode ) )
```
TheBrokenRail:
```suggestion
		std::string temp = m_worldsPath + '/' + de->d_name;
	
		struct stat buf;
		if ( ( lstat( temp.c_str(), &buf ) == 0 ) && S_ISDIR( buf.st_mode ) )
```

TheBrokenRail (changes requested):
Just some small suggestions.

iProgramMC:
I agree with TheBrokenRail's suggestions.

thirdparty/raknet/SocketLayer.cpp
SanyaSho: there's no use change it because raknet contains both ifdefs and if defineds

PR #98, opened 2023-11-05
Biome colour and grass side tint options

jonkadelic:
This PR adds:
- Support for biome-based leaf and grass tinting, using the grasscolor.png and foliagecolor.png from Beta 1.7.3. The option for this will be disabled if both of these files are not present.
- Support for tinting the sides of grass blocks, as was added in Beta 1.5. This uses a patch to add the transparent grass side texture to [4, 5] on terrain.png, as well as a flag called "grass_sides_tint". If grass_sides_tint is not set to true in patch_data.txt, the option to enable side tinting will be disabled.


iProgramMC:
Why not put `FoliageColor` and `GrassColor` into the client code? I mean the `World` project is shared between client and server (which doesn't yet exist but is going to at some point), and the server doesn't really need to know about colors now does it?

jonkadelic:
Because FoliageColor and GrassColor both need to be referenced in `Tile`, which is in shared code. If FoliageColor and GrassColor were moved to client code it would break server builds.

iProgramMC:
How about adding a `getTileColor` function into `TileRenderer` that returns either the advanced colors or the result of `Tile::getColor()`?

jonkadelic:
OK, I've moved them into client/renderer/ using that strategy.

iProgramMC:
BTW, the `c_` prefix in the patch data refers to Classic.

PR #109, opened 2023-12-30
Fix vs proj

JoeMama54:
Someone didn't fix the directories in the VS project file when they were updating the file structure.


iProgramMC:
Good catch, I'll run CI and then pull if everything is OK.

iProgramMC:
Nevermind, screw it.  CI doesn't check the VS projects either way

PR #110, opened 2023-12-31
SDL Center grabbed mouse & implement patch data

Stommm:
SDL didn't center grabbed mouse so it would always select random stuff in the hotbar. Added patch_data reading for the firework :-)

### DIFF
--- a/platforms/sdl/base/AppPlatform_sdl_base.cpp
+++ b/platforms/sdl/base/AppPlatform_sdl_base.cpp
@@ -26,6 +26,8 @@ void AppPlatform_sdl_base::_init(std::string storageDir, SDL_Window *window)
 	m_bShiftPressed[0] = false;
 	m_bShiftPressed[1] = false;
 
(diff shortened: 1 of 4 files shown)


iProgramMC:
Use SDL Read/Write operations. See here for an example: https://github.com/ReMinecraftPE/mcpe/blob/master/platforms/sdl/desktop/AppPlatform_sdl.cpp#L141

TheBrokenRail:
Rather than center the mouse, why not solve the root problem and prevent a grabbed mouse from clicking the hotbar? Just add a `if not mouse grabbed` to the relevant mouse handling function.

Stommm:
I don't like it when im in pause menu and the mouse is in some random place. Preventing hotbar from being clicked on pc is nice too

TheBrokenRail:
Personally, I'd prefer my mouse to be where it was before ut was grabbed rather than centered. This is just my opinion, though.

PR #113, opened 2024-01-16
iOS Support

BrentDaMage:
Remaking this pull-request since the last one was closed by iProgramMC accidentally.


BrentDaMage:
Ready for review

source/world/level/path/BinaryHeap.cpp
```diff
@@ -71,7 +71,7 @@ void BinaryHeap::downHeap(int num)
 		float var9;
 		if (var5 >= m_count) {
 			var8 = nullptr;
-			var9 = INFINITY;
+			var9 = std::numeric_limits<float>::infinity();
```
iProgramMC, line 74: Why was this changed?

iProgramMC, line 74: Why was this changed?

source/client/model/HumanoidModel.cpp
```diff
@@ -18,7 +18,9 @@ HumanoidModel::HumanoidModel(float a, float b):
 	m_leg1(0,  16),
 	m_leg2(0,  16)
 {
+	field_20 = false;
 	field_234 = 0;
+	field_20 = false;
```
iProgramMC, line 23: Remove the extraneous assignment

platforms/ios/minecraftpeAppDelegate.h
```diff
@@ -0,0 +1,28 @@
+//
+//  minecraftpeAppDelegate.h
+//  minecraftpe
+//
+//  Created by Brent on 10/12/23.
+//  Copyright (c) 2023 __MyCompanyName__. All rights reserved.
```
iProgramMC, line 6: Write "ReMinecraftPE" instead of `__MyCompanyName__` I guess.

source/world/level/path/BinaryHeap.cpp
iProgramMC: I know, it's because `INFINITY` was added shockingly late, in C++11.

BrentDaMage: In classic C++ fashion...

source/client/model/HumanoidModel.cpp
BrentDaMage: My bad, I moved that over from the header file without realizing it was already assigned at the top. Good catch.

PR #115, opened 2024-01-23
Fix RakNetSocket2 for iOS 6

BrentDaMage:


iProgramMC:
Yep, it works, thanks!

PR #116, opened 2024-01-24
Generate Missing Texture Texture

TheBrokenRail:
This generates a proper "missing texture" texture when one fails to load.

IMO this is way better than everything just defaulting to **black**. (Seriously, if you don't have assets loaded, the entire title-screen will be 100% black and it isn't that clear what the error is.)

This will also stop ReMCPE from spamming the console with error messages when it can't find a texture. (This issue was caused by it trying to load the texture every frame, but now it will only try to load it once.)

Screenshot:
Screenshot from 2024-01-23 22-18-53

### DIFF
--- a/source/client/renderer/Textures.cpp
+++ b/source/client/renderer/Textures.cpp
@@ -19,9 +19,19 @@ int Textures::loadTexture(const std::string& name, bool bIsRequired)
 
 	Texture t = m_pPlatform->loadTexture(name, bIsRequired);
 
-	int result = -1;
-	if (t.m_pixels)
-		result = assignTexture(name, t);
+	if (!t.m_pixels) {
+		t.field_C = 1;
+		t.field_D = 0;
+		t.m_width = 2;
+		t.m_height = 2;
+		t.m_pixels = new uint32_t[4];
+		t.m_pixels[0] = 0xfff800f8;
+		t.m_pixels[1] = 0xff000000;
+		t.m_pixels[3] = 0xfff800f8;
+		t.m_pixels[2] = 0xff000000;
+	}
+
+	int result = assignTexture(name, t);
 
 	return result;
 }


source/client/renderer/Textures.cpp
```diff
@@ -19,9 +19,19 @@ int Textures::loadTexture(const std::string& name, bool bIsRequired)
...
+		t.m_width = 2;
+		t.m_height = 2;
+		t.m_pixels = new uint32_t[4];
+		t.m_pixels[0] = 0xfff800f8;
+		t.m_pixels[1] = 0xff000000;
+		t.m_pixels[3] = 0xfff800f8;
+		t.m_pixels[2] = 0xff000000;
+	}
+
+	int result = assignTexture(name, t);
```
BrentDaMage: It looks like -1 was originally returned here to indicate an error occurred loading the texture. Surely there's a way to still allow for error detection & "texture not found" warnings while still keeping the placeholder textures.

TheBrokenRail: Since a texture is now always assigned, IMO, there's no reason for `Textures` to propagate the error.

Also, warnings still work. They're printed by `AppPlatform`.

BrentDaMage: As long as warnings still work, then I don't really see any downsides here. Looks all good!

iProgramMC: BTW, you should probably replace the duplicate `t.m_pixels` check with an if-else branching statement.

TheBrokenRail: What duplicate check?

iProgramMC: Oh! Sorry, I was looking in my mail and saw something that looked like this. Totally didn't see the `-` and `+` signs, sorry
image

source/client/renderer/Textures.cpp
```diff
@@ -19,9 +19,19 @@ int Textures::loadTexture(const std::string& name, bool bIsRequired)
 
 	Texture t = m_pPlatform->loadTexture(name, bIsRequired);
 
-	int result = -1;
-	if (t.m_pixels)
-		result = assignTexture(name, t);
+	if (!t.m_pixels) {
```
iProgramMC: If the texture could not be loaded, don't generate a missing texture if the bRequired parameter is false.

iProgramMC (changes requested):

source/client/renderer/Textures.cpp
TheBrokenRail: Done!

iProgramMC (approved):

PR #118, opened 2024-04-09
GUI: Add scrolling

### LINKED ISSUE #107: Can't scroll in the options menu
Just built for Linux, using scroll wheel, arrow keys, or page up/down keys, I can't seem to scroll down to the "Controls" section of the options screen. 
Screenshot_20231226_215539

er2off:
Fixes #107

### DIFF
--- a/source/client/gui/components/RolledSelectionList.cpp
+++ b/source/client/gui/components/RolledSelectionList.cpp
@@ -335,3 +335,10 @@ void RolledSelectionList::clickedHeader(int x, int y)
 {
 
 }
+
+void RolledSelectionList::handleScroll(bool down)
+{
+	float diff = 5.0f * (down ? -1.0f : 1.0f);
+	field_34 = field_30 = field_30 + diff;
+	field_28 = 0;
+}
--- a/source/client/gui/components/ScrolledSelectionList.cpp
+++ b/source/client/gui/components/ScrolledSelectionList.cpp
@@ -97,7 +97,6 @@ void ScrolledSelectionList::renderScrollBackground()
 
 void ScrolledSelectionList::checkInput(int mouseX, int mouseY)
 {
-	int nItems = getNumberOfItems();
 	if (Mouse::isButtonDown(BUTTON_LEFT))
 	{
 		if (float(mouseY) >= field_C && float(mouseY) <= field_10 && abs(mouseY - field_28) > 5)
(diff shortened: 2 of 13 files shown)


iProgramMC:
Although stuttering occurs when scrolling through the world select list, it also does when dragging, so this isn't a problem with this PR.

source/client/gui/components/ScrolledSelectionList.cpp
```diff
@@ -276,3 +275,10 @@ void ScrolledSelectionList::setRenderHeader(bool b, int i)
 		i = 0;
 	field_48 = i;
 }
+
+void ScrolledSelectionList::handleScroll(bool down)
+{
+	float diff = 5.0f * (down ? -1.0f : 1.0f);
```
iProgramMC, line 281: Feels weird.  Like, when you scroll up, and then down, it seems to "continue" scrolling in the direction I was first scrolling in. Maybe it should just instantly set the velocity.

Also I think you should add a timeout of a few ms where the world select is just loose (as if you were holding a finger on the scroll area)

iProgramMC (changes requested):
GitHub forces me to write something here as well.  All's good, except for that one issue I just mentioned.

source/client/gui/components/ScrolledSelectionList.cpp
er2off: This comment is for RolledSelectionList? Problem doesn't appear with ScrolledSelectionList.

iProgramMC: Yes, oops... Comment is related to RolledSelectionList.

source/client/gui/components/RolledSelectionList.cpp
```diff
@@ -335,3 +335,10 @@ void RolledSelectionList::clickedHeader(int x, int y)
 {
 
 }
+
+void RolledSelectionList::handleScroll(bool down)
+{
+	float diff = 5.0f * (down ? -1.0f : 1.0f);
+	field_34 = field_30 = field_30 + diff;
+	field_28 = 0;
```
uniformization, line 343: you should probably rename the unnamed fields that you use, only a suggestion though

er2off: Agree with this but I think these classes should be rewritten instead.

iProgramMC: > Agree with this but I think these classes should be rewritten instead.

Disagree. These classes implement the base of scrollable list views. I don't see how they need to be rewritten.

PR #122, opened 2024-04-30
Add isometric screenshot feature.

iProgramMC:


source/client/app/Minecraft.cpp
```diff
@@ -503,9 +503,14 @@ void Minecraft::tickInput()
...
-				Minecraft::useAmbientOcclusion = getOptions()->m_bAmbientOcclusion;
-				m_pLevelRenderer->allChanged();
+				// getOptions()->m_bAmbientOcclusion = !getOptions()->m_bAmbientOcclusion;
+				// Minecraft::useAmbientOcclusion = getOptions()->m_bAmbientOcclusion;
+				// m_pLevelRenderer->allChanged();
+
+				if (m_pGameRenderer->m_bIsometric)
+					m_pGameRenderer->m_isomStage = 100000;
+				else
+					m_pGameRenderer->startIsometricRender();
```
er2off: uhm... no ambient occlusion toggle? maybe remap to other button?

source/client/renderer/GameRenderer.cpp
```diff
@@ -680,7 +863,7 @@ void GameRenderer::render(float f)
 	debugText << "ReMinecraftPE " << m_pMinecraft->getVersionString();
 	debugText << "\n" << m_shownFPS << " fps, " << m_shownChunkUpdates << " chunk updates";
 
-	if (m_pMinecraft->getOptions()->m_bDebugText)
+	if (m_pMinecraft->getOptions()->m_bDebugText && !m_pMinecraft->getOptions()->m_bDontRenderGui)
```
er2off, line 889: finally! actually it's not part of isometric renderer but anyway

source/client/renderer/GameRenderer.cpp
```diff
@@ -542,12 +593,142 @@ void GameRenderer::renderLevel(float f)
...
+{
+	const float isomSizeX = C_ISOM_TILES_PER_WIDTH(float(Minecraft::width) / float(Minecraft::height));
+	const float isomSizeY = C_ISOM_TILES_PER_HEIGHT;
+	const int isomWidth = int(C_MAX_CHUNKS_X * 16 / isomSizeX / 2) + 2;
+	const int isomHeight = int(C_MAX_CHUNKS_X * C_MAX_CHUNKS_Z / isomSizeY) + 2;
+	const int maxIsomStage = isomWidth * isomHeight;
+
+	if (m_bIsometric) {
+		LOG_I("Saving %d of %d", m_isomStage, maxIsomStage);
+		std::string fn = "IsomShots\\" + std::to_string(m_isomStage) + ".png";
```
er2off: windows detected!

source/client/app/Minecraft.cpp
iProgramMC: This is temporary and will be replaced.

source/client/renderer/GameRenderer.cpp
iProgramMC: To be honest this condition is redundant, as if `m_bDontRenderGui` is true, there's an early return

source/client/renderer/GameRenderer.cpp
iProgramMC: This is also temporary, and will be replaced.

source/client/renderer/GameRenderer.cpp
```diff
@@ -641,8 +845,10 @@ void GameRenderer::render(float f)
 			renderLevel(f);
 			if (m_pMinecraft->getOptions()->m_bDontRenderGui)
 			{
-				if (!m_pMinecraft->m_pScreen)
+				if (!m_pMinecraft->m_pScreen) {
```
er2off, line 848: We definitely need to set up something like clang-format, code is using Microsoft-ish style but some code like this formatted incorrectly.

PR #125, opened 2024-05-13
Double tap to fly

kylegg:
- Allows flight to be toggled when KM_FLY_UP is double tapped
- Flight keys x and c were swapped out for space and shift


iProgramMC:
Flight is currently a bit of a hack because you can no-clip. I don't knownhow good it would be to allow that without going to options. Perhaps we should disable noclip in flight mode.

kylegg:
Sounds like something that would be nice. Sneaking seems to interfere with flight when collision is turned on so I'm going to have to do a bit more tinkering to fix that before I can tack on another commit.

PR #127, opened 2024-06-18
The Big CMake + MinGW + Sound PR

### LINKED ISSUE #119: Load Sounds From Filesystem
Currently sounds are just embedded in memory. We should update our audio API abstraction layers to support loading sounds from files. Streaming sounds from files would also allow us to add in-game music.

This means updating:
1. SoundSystemAL
2. SoundSystemDS
3. SoundSystemSL

TheBrokenRail:
This PR:
- Refactors the CMake build-system:
  - There is now a top-level `CMakeLists.txt` so you no longer need to `cd platforms/sdl` before building the project.
  - SDL is now loaded in `platforms/sdl/CMakeLists.txt` like you would expect instead of in `source/CMakeLists.txt`.
  - Sound systems have been decoupled from SDL/Win32. You can now build SDL with DirectSound or Win32 with OpenAL.
  - The Win32 platform can now be built with CMake using MinGW-w64.
  - The SDL platform can also now be built with CMake using MinGW-w64.
  - Both of those configurations have been added to the CI.'
- Sound is now loaded at runtime instead of being loaded from the binary! (Fixes #119)
  - `tools/grabsounds.py` now extracts sounds to `game/assets/sound` using [LIEF](https://github.com/lief-project/LIEF) (which is downloaded automatically).
  - Sound data is loaded using `AppPlatform::readAssetFile`, which has been implemented (and tested) for both SDL, Win32, and native Android.
  - `AppPlatform::getPatchData` has been changed to use `AppPlatform::readAssetFile`.
- Extra Notes:
  - The macOS build currently only works because I added a `ln -s` to the CI, someone should probably fix that.
  - I also adjusted the line width calculation for block outlines.
  - And I updated the [GLES Compatibility Layer](https://gitea.thebrokenrail.com/minecraft-pi-reborn/gles-compatibility-layer).
  - Finally, I fixed some bugs with `Logger`.


source/client/sound/SoundRepository.cpp
```diff
@@ -21,7 +21,7 @@ void SoundRepository::add(const std::string& name, SoundDesc& sd)
 	{
 		std::vector<SoundDesc> sdv;
 		sdv.push_back(sd);
-		m_repo.insert(std::pair<std::string, std::vector<SoundDesc>>(name, sdv));
+		m_repo.insert(std::pair<std::string, std::vector<SoundDesc> >(name, sdv));
```
iProgramMC: Use `std::make_pair`.

platforms/sdl/main.cpp
```diff
@@ -305,6 +316,10 @@ void CheckOptionalTextureAvailability()
 // Main
 int main(int argc, char *argv[])
 {
+	// Setup Logging
+	new Logger;
```
iProgramMC: You aren't assigning a pointer.  Does `Logger`'s constructor register that instance anywhere? Frankly the constructor shouldn't take care of any of that

platforms/sound/directsound/CustomSoundSystem.hpp
```diff
@@ -49,4 +49,6 @@ class SoundSystemDS : public SoundSystem
 	IDirectSound* m_directsound;
 	LPDIRECTSOUND3DLISTENER m_listener;
 	std::vector<BufferInfo> m_buffers;
-};
\ No newline at end of file
+};
+
+#define SOUND_SYSTEM SoundSystemDS
```
iProgramMC, line 57: I disagree with renaming all sound systems to `CustomSoundSystem`. I especially disagree with your choice of using a `#define` where a `typedef` would have fit incredibly well.

I would say revert this and then in each AppPlatform, use preprocessor conditions to determine the sound system *in that AppPlatform*. And use `typedef` for that.

You can't really use DirectSound on non-Windows; nor can you use OpenSL ES on desktop Linux and Windows.

platforms/windows/main.cpp
```diff
@@ -142,6 +143,10 @@ int WINAPI WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPSTR lpCmdLine
 
 	SetInstance(hInstance);
 
+	// This initializes the Logger singleton to use the Windows-specific variant
+	// If we didn't initialize it here, the Minecraft class would have our back
+	new LoggerWin32;
```
iProgramMC: As said before, I disagree with this way of thinking.  I'd say something like `Logger::setSingleton(new LoggerWin32)` would work best.

source/client/app/AppPlatform.cpp
```diff
@@ -210,7 +210,15 @@ bool AppPlatform::hasFileSystemAccess()
 
 std::string AppPlatform::getPatchData()
 {
-	return "";
+	AssetFile file = readAssetFile("patches/patch_data.txt");
+	std::string out = std::string(file.data, file.data + file.size);
+	delete file.data;
```
iProgramMC, line 231: You should delete `data` within the destructor and add move and copy constructors to `AssetFile`. Because we're using C++.

source/client/app/NinecraftApp.cpp
```diff
@@ -57,7 +57,7 @@ void NinecraftApp::initGLStates()
 	glCullFace(GL_BACK);
 	glEnable(GL_TEXTURE_2D);
 	glHint(GL_PERSPECTIVE_CORRECTION_HINT, GL_FASTEST);
-	glDisable(GL_LIGHTING);
+	//glDisable(GL_LIGHTING);
```
iProgramMC: Why?

source/client/gui/components/RolledSelectionList.cpp
```diff
@@ -163,7 +163,7 @@ void RolledSelectionList::render(int mouseX, int mouseY, float f)
 	
 	capXPosition();
 	
-	glDisable(GL_LIGHTING);
+	//glDisable(GL_LIGHTING);
```
iProgramMC: Why?

source/client/gui/components/ScrolledSelectionList.cpp
```diff
@@ -161,7 +161,7 @@ void ScrolledSelectionList::render(int mouseX, int mouseY, float f)
 	field_38 *= 0.75f;
 	capYPosition();
 
-	glDisable(GL_LIGHTING);
+	//glDisable(GL_LIGHTING);
```
iProgramMC: Why?

source/client/gui/screens/StartMenuScreen.cpp
```diff
@@ -13,7 +13,7 @@
 #include "SelectWorldScreen.hpp"
 #include "JoinGameScreen.hpp"
 
-#if defined(_WIN32) || (defined(TARGET_OS_MAC) && TARGET_OS_IPHONE == 0)
+#if defined(USE_SDL) || defined(_WIN32) || (defined(TARGET_OS_MAC) && TARGET_OS_IPHONE == 0)
```
iProgramMC: `USE_SDL` might be defined on Android.  Now, I don't know why Android games don't give you a quit option, but it's probably best to imitate them.

source/client/renderer/GameRenderer.cpp
```diff
@@ -329,7 +329,7 @@ void GameRenderer::setupFog(int i)
 	fog_color[3] = 1.0f;
 
 	glFogfv(GL_FOG_COLOR, fog_color);
-	glNormal3f(0.0f, -1.0f, 0.0f);
+	//glNormal3f(0.0f, -1.0f, 0.0f);
```
iProgramMC: Why?

source/client/renderer/ItemInHandRenderer.cpp
```diff
@@ -187,7 +187,7 @@ void ItemInHandRenderer::render(float f)
 		glTranslatef(-0.3f * Mth::sin(float(M_PI) * Mth::sqrt(fAnim)), 0.4f * Mth::sin(2.0f * float(M_PI) * Mth::sqrt(fAnim)), -0.4f * Mth::sin(float(M_PI) * fAnim));
 		glTranslatef(0.64f, ((1.0f - f1) * -0.6f) - 0.6f, -0.72f);
 		glRotatef(45.0f, 0.0f, 1.0f, 0.0f);
-		glEnable(GL_RESCALE_NORMAL);
+		//glEnable(GL_RESCALE_NORMAL);
```
iProgramMC: Why?

source/client/renderer/LevelRenderer.cpp
```diff
@@ -456,7 +456,7 @@ void LevelRenderer::render(Mob* pMob, int a, float b)
 				c = m_chunksLength;
 
 			glDisable(GL_TEXTURE_2D);
-			glDisable(GL_LIGHTING);
+			//glDisable(GL_LIGHTING);
```
iProgramMC: Why?

source/client/renderer/LevelRenderer.cpp
```diff
@@ -811,17 +811,21 @@ void LevelRenderer::renderHitOutline(Player* pPlayer, const HitResult& hr, int i
...
+	if (range[1] < line_width)
+	{
+		line_width = range[1];
+	}
+	else if (range[0] > line_width)
+	{
+		line_width = range[0];
+	}
+	// Set Line Width
+	glLineWidth(line_width);
```
iProgramMC, line 925: Why switch to aliased lines?

source/client/renderer/entity/ItemRenderer.cpp
```diff
@@ -52,7 +52,7 @@ void ItemRenderer::render(Entity* pEntity, float x, float y, float z, float a, f
 		itemsToRender = 4;
 
 	glTranslatef(x, y + 0.1f + yOffset * 0.1f, z);
-	glEnable(GL_RESCALE_NORMAL);
```
iProgramMC: Why?

source/client/renderer/entity/ItemSpriteRenderer.cpp
```diff
@@ -10,7 +10,7 @@ void ItemSpriteRenderer::render(Entity* pEntity, float x, float y, float z, floa
 {
 	glPushMatrix();
 	glTranslatef(x, y, z);
-	glEnable(GL_RESCALE_NORMAL);
```
iProgramMC: Why?

source/client/renderer/entity/MobRenderer.cpp
```diff
@@ -176,7 +176,7 @@ void MobRenderer::renderNameTag(Mob* mob, const std::string& str, float x, float
 
 	glPushMatrix();
 	glTranslatef(x + 0.0f, y + 2.3f, z);
-	glNormal3f(0.0f, 1.0f, 0.0f);
```
iProgramMC: Why?

source/client/sound/SoundData.cpp
```diff
@@ -8,13 +8,35 @@
...
+void SoundDesc::_load(const AppPlatform* platform)
+{
+#define SOUND(category, name) SA_##name._load(platform, #name);
+#include "sound_list.h"
+#undef SOUND
+}
+
+#define SOUND(category, name) SoundDesc SA_##name;
+#include "sound_list.h"
+#undef SOUND
```
iProgramMC: How does this work? You need to put the declarations above any of their uses, don't you?

source/client/sound/SoundRepository.cpp
```diff
@@ -12,6 +12,10 @@
 
 void SoundRepository::add(const std::string& name, SoundDesc& sd)
 {
+	if (!sd.isLoaded)
+	{
```
iProgramMC: Remove brackets here

source/client/sound/SoundData.cpp
```diff
@@ -8,13 +8,35 @@
...
+
+void SoundDesc::_load(const AppPlatform* platform, const char *name)
+{
+	if (isLoaded)
+	{
+		// Already Loaded
+		return;
+	}
+	// Read Asset File
+	AssetFile file = platform->readAssetFile(std::string("sound/") + name + ".pcm");
```
iProgramMC: I suggest replacing periods with slashes here. So it'll try to access, for example, `sound/step/gravel.pcm` instead of `sound/step.gravel.pcm`.

source/common/Logger.cpp
```diff
@@ -15,8 +15,11 @@ Logger* const Logger::singleton()
 Logger::Logger()
 {
     // Stick with the first output handle we get
-    if (!m_singleton)
```
iProgramMC, line 18: I commented on this earlier. I suggest you use a single `Logger::setSingleton` function and call it like `Logger::setSingleton(new LoggerWhatever);`

iProgramMC (changes requested):

iProgramMC:
Also worth noting that I don't really like giant PRs. Which is why I ask that you create separate PRs for different changes from now on.

platforms/sdl/main.cpp
TheBrokenRail: > Does `Logger`'s constructor register that instance anywhere?

Yup.

Ideally the singleton would be initialized in `Logger.cpp`. The issue is the this would prevent the Win32 code from allocating its own `LoggerWin32`. So it starting the logger must be the responsibility of `main`. It's ugly, but it works without double-allocating or double-freeing anything.

platforms/sound/directsound/CustomSoundSystem.hpp
TheBrokenRail: > You can't really use DirectSound on non-Windows; nor can you use OpenSL ES on desktop Linux and Windows.

The issue is OpenAL. It can be used on both SDL and Win32, so you end up with:
```c++
#ifdef USE_OPENAL
#include "SoundSystemAL.h"
#define SOUND_SYSTEM SoundSystemAL
#else
#include "SoundSystemDS.h"
#define SOUND_SYSTEM SoundSystemDS
#endif
```
In both `platforms/sdl/main.cpp` and `platforms/windows/main.cpp`.

I'll switch to `typedef`s though, that is nicer.

platforms/windows/main.cpp
TheBrokenRail: I'll switch to that.

source/client/app/AppPlatform.cpp
TheBrokenRail: Wouldn't that require `SoundDesc` maintaining a copy of `AssetFile` so the sound data isn't deleted?

source/client/app/NinecraftApp.cpp
TheBrokenRail: I removed support for GL lighting in my GLES compatibility layer because literally nothing was using it and removing it simplified the code (and shaders) a lot. In this case, `GL_LIGHTING` is disabled a lot but never actually enabled or configured.

I can easily add it back if anything actually needs it.

source/client/renderer/LevelRenderer.cpp
TheBrokenRail: Because smoothed lines aren't actually supported on OpenGL ES 2.0. It was just failing silently.

source/client/sound/SoundData.cpp
TheBrokenRail: Yup, it's an [X-macro](https://en.wikipedia.org/wiki/X_macro?useskin=vector).

source/client/sound/SoundData.cpp
TheBrokenRail: There are no periods. It just accesses `gravel1.pcm`.

source/client/sound/SoundData.cpp
iProgramMC: I know what an x-macro is.  [In fact I've used them before](https://github.com/iProgramMC/NanoShellOS/blob/master/src/icons.h).  I'm talking about how it works when none of the SA_XYZ variables are defined when used right above.

TheBrokenRail:
Yeah, sorry about that. I got a bit carried away (clearly).

platforms/sdl/main.cpp
iProgramMC: No but I mean, the constructor shouldn't be the one assigning the singleton, it looks weird and like a memory leak here. (Even though it's not)

platforms/sound/directsound/CustomSoundSystem.hpp
iProgramMC: Yep, that's fine. I don't mind that at all.

source/client/app/AppPlatform.cpp
iProgramMC: Yes. But you return `AssetFile` from `AppPlatform::readAssetFile` so you can just keep the `AssetFile` in memory (ideally you'd also use `std::move` to avoid copying)

source/client/sound/SoundData.cpp
iProgramMC: Didn't catch that, sorry.  I find it kind of weird that there are multiple sounds with the same in-game ID, but it's probably just to support random sounds easily.

source/client/app/NinecraftApp.cpp
iProgramMC: I suppose just disable it once in `initGLStates` then, in the case that the OpenGL version provided does support lighting. As in,
```cpp
#ifndef USE_GLES1_COMPATIBILITY_LAYER
    glDisable(GL_LIGHTING);
#endif
```

TheBrokenRail: But why? It's never enabled and disabled by default

iProgramMC: Great! Then this is resolved

For reference (source: https://registry.khronos.org/OpenGL-Refpages/gl2.1/xhtml/glLight.xml):
> To enable and disable lighting calculation, call [glEnable](https://registry.khronos.org/OpenGL-Refpages/gl2.1/xhtml/glEnable.xml) and [glDisable](https://registry.khronos.org/OpenGL-Refpages/gl2.1/xhtml/glDisable.xml) with argument GL_LIGHTING. Lighting is initially disabled.

source/client/sound/SoundData.cpp
TheBrokenRail: They're also defined in `SoundData.hpp`:
```c++
#define SOUND(category, name) extern SoundDesc SA_##name;
#include "sound_list.h"
#undef SOUND
```

source/client/app/AppPlatform.cpp
TheBrokenRail: So, funny story. Turns out `SoundRepository` copies `SoundDesc` *a lot*.

So the options are:
- Modify `SoundRepository` to store pointers or references (which in turn would require modifying the `SoundSystem`s).
- Add a copy constructor to `AssetFile` and copy the sound data whenever it is used, which sounds... really inefficient.
- Keep the data externally-managed.

source/client/renderer/GameRenderer.cpp
TheBrokenRail: This doesn't do anything without `GL_LIGHTING`, and that is never enabled.

source/client/renderer/ItemInHandRenderer.cpp
TheBrokenRail: This doesn't do anything without `GL_LIGHTING`, and that is never enabled.

source/client/sound/SoundRepository.cpp
TheBrokenRail: Done!

platforms/sdl/main.cpp
TheBrokenRail: Fixed.

platforms/windows/main.cpp
TheBrokenRail: Fixed.

source/client/gui/screens/StartMenuScreen.cpp
TheBrokenRail: Fixed.

source/client/sound/SoundRepository.cpp
TheBrokenRail: Done.

source/common/Logger.cpp
TheBrokenRail: Done.

TheBrokenRail:
Also, I added `tools/extract_apk.py`. It just calls `grabsounds,py` and extracts the other assets to `game/assets`. I mainly created it because I got annoyed with having to figure out how to merge-copy the game assets into `game/assets`. It also shows a file chooser if you run it by itself, which should make it easier to use for beginners.

TheBrokenRail:
When I get more time, I'm going to split this up into multiple PRs. Until then, I'll mark it as WIP.

BrentDaMage:
We should really be loading all of our SoundDesc structs dynamically based on the file hierarchy inside /assets/sounds. MCPE 0.12.1 has an "example" of this, even though it's getting most of its info from a JSON file.

BrentDaMage:
@TheBrokenRail Could you please fix the CMake files? Not sure why the SDL2 and GLES compatibility layer stuff is throwing back errors. I'm gonna make sure the Xcode project works for macOS and iOS tomorrow.

BrentDaMage:
Additionally, it's worth nothing, that for whatever reason, an Ogg Vorbis decoder is already present in one of our thirdparty libraries in ~/thirdparty/stb_image/include/stb_vorbis.c

iProgramMC:
> for whatever reason
Yes, because we clone this repo https://github.com/nothings/stb in stb_image. The name of that submodule is actually a misnomer as we import the entirety of stb, not just stb_image.

TheBrokenRail:
CMake changes are in #149.

BrentDaMage:
For clarification, are sound changes also now being split out into a new PR as well? If so, is this PR going to be archived/closed?

TheBrokenRail:
Yeah, once I make the sound PR, I'm closing this one.

TheBrokenRail:
This branch is a disaster.

BrentDaMage:
> Yeah, once I make the sound PR, I'm closing this one.

Cool. I'm planning on adding OGG support on my flight back to the U.S.
I'm hoping to not have to deal with merge conflict hell.

TheBrokenRail:
The sound code has been split off into its own PR: #150.

PR #132, opened 2024-08-22
Add /give & /clear command

morgana-x:
/give [item] [count]
+ Only works for the host

/clear
+ Only works for the host
+ Clears player inventory


source/network/ServerSideNetworkHandler.cpp
```diff
@@ -520,3 +522,49 @@ void ServerSideNetworkHandler::commandTP(OnlinePlayer* player, const std::vector
...
+	int amount = 1;
+	if (!sscanf(parms[0].c_str(), "%d", &id) || !sscanf(parms[1].c_str(), "%d", &amount))
+	{
+		sendMessage(player, "Usage: /give [item] [item count]");
+		return;
+	}
+
+	Inventory* pInventory = player->m_pPlayer->m_pInventory;
+
+	pInventory->addTestItem(id, amount);
```
Bigjango13: Couldn't you check if `Item::items[id]` is null before this to prevent crashes? Although I'm not sure if Item::items is zeroed beforehand in MCPE (in MCPI it mostly is, except for a single special case, and in MCJE it is).

morgana-x: I'll try that when I'm back home in about 4 hours, thankyou

morgana-x: Added a check for that as well as if the id is greater than 512 or less than 0, resolving the possibility of a crash

morgana-x:
Closed pull request due to accidentally committing unrelated stuff to the main branch of my fork

PR #134, opened 2024-08-27
Several fixes and improvements...

KoopaOnGit:
* Added ability to arbritrarily block/unblock keyboard input at any time, fixes a 'white screen' issue when pressing ESC right at the time where the save screen is shown - on Linux this even segfaults for some reason, haven't investigated root cause but this definitely seems to reliably stop it from happening on a higher level.

* Usage of SAFE_DELETE when SAFE_DELETE_ARRAY should be used for the sake of correctness

* Zero-init float arrays in FireTexture constructors

* Fix OOB read during Font::init when initializing a Font that may not exist or has corrupted data (both VS-ASan and Dr.Memory complained about it)

### DIFF
--- a/source/client/app/Minecraft.cpp
+++ b/source/client/app/Minecraft.cpp
@@ -1163,6 +1163,8 @@ int Minecraft::getFpsIntlCounter()
 
 void Minecraft::leaveGame(bool bCopyMap)
 {
+	Keyboard::setBlocked(true);
+
 	m_bPreparingLevel = false;
 	m_pRakNetInstance->disconnect();
 	m_pMobPersp = nullptr;
--- a/source/client/gui/screens/StartMenuScreen.cpp
+++ b/source/client/gui/screens/StartMenuScreen.cpp
@@ -365,7 +365,7 @@ StartMenuScreen::StartMenuScreen() :
 
 StartMenuScreen::~StartMenuScreen()
 {
-	SAFE_DELETE(m_pTiles);
+	SAFE_DELETE_ARRAY(m_pTiles);
 }
 
 void StartMenuScreen::_updateLicense()
--- a/source/client/player/input/Keyboard.cpp
+++ b/source/client/player/input/Keyboard.cpp
@@ -13,6 +13,7 @@
 std::vector<KeyboardAction> Keyboard::_inputs;
 int Keyboard::_index = -1;
 Keyboard::KeyState Keyboard::_states[KEYBOARD_STATES_SIZE];
+bool Keyboard::_blocked = false;
 
 void Keyboard::feed(KeyState state, int key)
 {
@@ -30,7 +31,7 @@ void Keyboard::feed(KeyState state, int key)
 
 bool Keyboard::next()
 {
-	if (_index + 1 >= _inputs.size())
+	if (_blocked || _index + 1 >= _inputs.size())
 		return false;
 
 	_index++;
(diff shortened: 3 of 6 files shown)


source/client/app/Minecraft.cpp
```diff
@@ -1163,6 +1163,8 @@ int Minecraft::getFpsIntlCounter()
 
 void Minecraft::leaveGame(bool bCopyMap)
 {
+	Keyboard::setBlocked(true);
```
TheBrokenRail, line 1212: When is the keyboard unblocked?

KoopaOnGit: Keyboard::reset() unblocks it - So, no worries...

TheBrokenRail:
IMO, instead of just disabling the keyboard arbitrarily while leaving the world, it would be better to determine (and fix) the actual bugs instead of just hiding them.

KoopaOnGit:
You can do so by getting rid of the keyboard blocking code and reproducing the bug - but to be honest, I couldn't debug the root issues as I remember because:

1. On Windows it only gives out a weird white screen and I have no prior experience towards debugging something like this as I lack knowledge of how to trace back rendering related issues, or I did just really try and did not want to bother.

2. The only linux platform where I have ReMCPE working is the raspberry pi, and on there it actually has a segfault when you try to do what I stated above, yet debugging and applying patches on the Pi is really tedious so I didn't really bother either...


I think I'll try to assess the root causes and uncomment it myself and provide a patch here too now, as I guess I'll really try this time.

iProgramMC:
> I identified the root cause image
> 
> Commenting this out and the issue no longer persists. But here's the question: What is this even used for?

This is used to exit out of menus, such as the pause menu, or inventory menu.  I believe you have to override the `Screen::keyPressed(int)` method in the SavingWorldScreen to prevent the white screen issue.

iProgramMC:
Can you put the following changes in a new PR?

- The fix to change `SAFE_DELETE(m_pTiles)` to `SAFE_DELETE_ARRAY(m_pTiles)` in `StartMenuScreen.cpp`
- The fix to change `SAFE_DELETE(m_data1)` to `SAFE_DELETE_ARRAY(m_data1)` in `FireTexture.cpp` (same for `m_data2`)

The keyboard block functionality I completely disagree with. The underlying bugg can just be fixed.

https://github.com/ReMinecraftPE/mcpe/pull/134/files#diff-b6f242667aac4d5605d75f5cfc05d95bc2b2487299360c9c7bb7a10bfc3148aaR42-R48
The font code should be rearchitected to scale with the texture size.

iProgramMC:
> Or just block all keyboard input during those scenes where any input isn't supposed to be handled anyway and call it a day?

Sure but this is just a bandage fix.  Simply fix the root cause and we're good to go.

iProgramMC:
I said that the font texture width generation needs to be rearchitected already.

source/client/gui/Screen.cpp
```diff
@@ -71,10 +71,11 @@ bool Screen::isInGameScreen()
 
 void Screen::keyPressed(int key)
 {
-	if (m_pMinecraft->getOptions()->isKey(KM_MENU_CANCEL, key))
+	//@BUG: Why set the screen to nullptr here? Causes white-screen issue when hitting ESC during world save menu...
```
BrentDaMage: Setting the screen to nullptr makes it so no screen is displayed, effectively displaying whatever is going on in-game. This was the intended behavior, and was necessary to exit the pause menu. It has since been replaced with `NinecraftApp::handleBack(bool)`.

BrentDaMage:
Are the issues this PR is meant to fix still present?

PR #136, opened 2024-09-15
Support C++ 11 for xbox 360 and add fakegl

Wolf3s:


iProgramMC:
Dude... just define USE_OLD_CPP (or don't) when targeting Xbox 360.

Wolf3s:
> Dude... just define USE_OLD_CPP (or don't) when targeting Xbox 360.
> 
> Anyways FakeGL seems good for a temporary solution, but @BrentDaMage is working on a [graphics interface](https://github.com/BrentDaMage/mcpe-engine) that will replace OpenGL's use in ReMCPE (and be implemented using OpenGL, DirectX etc)

I will see about it msvc supports constexpr if use it as a macro.

PR #147, opened 2025-04-13
Fix zlib build issue created by Xcode 16.3

BrentDaMage:


BrentDaMage:
This still needs to be tested on an old version of Xcode (around Mac OS X Snow Leopard).
This fixes build issues on for the latest version of Xcode, and I'd like to make sure it doesn't break anything on legacy Xcode versions.

BrentDaMage:
PR tested. Does not appear to cause build issues on Mac OS X Lion.

PR #149, opened 2025-04-20
MinGW Support + CMake Refactor

TheBrokenRail:
This PR:
* Splits sound backends from platform backends
  * This allows you to easily build Android with OpenSL or OpenAL
  * Or build Windows with OpenAL or DirectSound
* Adds support for building the CMake project with MinGW-w64
  * This is added to the CI
  * This allows the Windows version to be built on Linux
* Changes `Logger` to be initialized in `main` rather than `AppPlatform`
  * This fixes some bugs where messages would fail to log after the `AppPlatform` is destroyed
* Finally removes the GLES comparability layer
* Enables OpenGL lighting on all platforms
* Fixes other miscellaneous bugs


platforms/sdl/base/AppPlatform_sdl_base.cpp
```diff
@@ -172,16 +173,6 @@ int AppPlatform_sdl_base::getScreenHeight() const
 
 void AppPlatform_sdl_base::recenterMouse()
 {
-	// Note. The only reason we do it this way instead of
```
BrentDaMage, line 175: This is necessary to center the mouse whenever a Screen is opened. As discussed, this breaks WASM, so please move to `AppPlatform_sdl`.

BrentDaMage (approved):
LGTM

platforms/sdl/base/AppPlatform_sdl_base.cpp
TheBrokenRail: ...I already did?

source/client/renderer/entity/MobRenderer.cpp
```diff
@@ -213,7 +213,9 @@ void MobRenderer::renderNameTag(Mob* mob, const std::string& str, float x, float
 
 	glPushMatrix();
 	glTranslatef(x + 0.0f, y + 2.3f, z);
+#ifndef __EMSCRIPTEN__
```
BrentDaMage, line 216: Why not use `USE_GL_NORMAL_LIGHTING`?

TheBrokenRail: Because while Emscripten does support OpenGL lighting (hence it has `USE_GL_NORMAL_LIGHTING` defined), its implementation of `glNormal3f` is [extremely incomplete](https://github.com/emscripten-core/emscripten/issues/24153). Also, side-note, I'm pretty sure that function call doesn't do anything anyway.

BrentDaMage: Fair enough.

PR #150, opened 2025-04-21
Load Sound Data From Files

### LINKED ISSUE #119: Load Sounds From Filesystem

TheBrokenRail:
Fixes #119

This PR:
* Modifies ReMCPE to load sound data from either `.pcm` or `.ogg` files instead of embedding it in the binary
* Modifies the Python sound extractor to generate `.pcm` files
  * Also adds a Python scripts that extracts an APK, including both textures and sounds


platforms/windows/projects/Client/Client.vcxproj
```diff
@@ -257,10 +261,12 @@
     <ClCompile Include="$(MC_ROOT)\source\client\renderer\TileRenderer.cpp" />
     <ClCompile Include="$(MC_ROOT)\source\client\renderer\WaterSideTexture.cpp" />
     <ClCompile Include="$(MC_ROOT)\source\client\renderer\WaterTexture.cpp" />
+    <ClCompile Include="..\..\..\..\source\client\sound\SoundPathRepository.cpp" />
```
TheBrokenRail: Shouldn't this include `$(MC_ROOT)`?

platforms/windows/projects/Client/Client.vcxproj.filters
```diff
@@ -284,6 +284,9 @@
     <ClInclude Include="$(MC_ROOT)\source\client\renderer\VertexPT.hpp">
       <Filter>Header Files\Renderer</Filter>
     </ClInclude>
+    <ClInclude Include="..\..\..\..\source\client\sound\SoundPathRepository.hpp">
```
TheBrokenRail: Same here.

platforms/windows/projects/Client/Client.vcxproj.filters
```diff
@@ -664,6 +676,9 @@
     <ClCompile Include="$(MC_ROOT)\source\client\renderer\WaterSideTexture.cpp">
       <Filter>Source Files\Renderer</Filter>
     </ClCompile>
+    <ClCompile Include="..\..\..\..\source\client\sound\SoundPathRepository.cpp">
```
TheBrokenRail: And here.

source/world/entity/Mob.cpp
```diff
@@ -201,96 +202,108 @@ void Mob::tick()
...
-		m_rotPrev.y += 360.0f;
+	while (m_rot.y - m_oRot.y >= 180.0f)
+		m_oRot.y += 360.0f;
 
 	field_B54 += x2;
 }
 
 void Mob::baseTick()
 {
-	m_oAttackAnim = m_attackAnim;
```
TheBrokenRail, line 227: A good chunk of changes to this file seem to solely be indentation changes. These should probably be changed back to tabs, unless the project's code-style is being changed.

platforms/sound/openal/SoundStreamAL.cpp
```diff
@@ -0,0 +1,178 @@
...
+}
+
+bool SoundStreamAL::_stream(uint32_t buffer)
+{
+    int size = 0;
+    int result = 0;
+
+    while (size < _tempPcmBufferSize)
+    {
+        result = stb_vorbis_get_samples_short_interleaved(_decoder, _info.channels, _tempPcmBuffer + size, _tempPcmBufferSize - size);
```
TheBrokenRail: Shouldn't OGG decoding be in platform-agnostic code?

source/world/entity/Mob.cpp
BrentDaMage: There is no code-styling dictated for spaces vs tabs. There are actually tons of places where they are inconsistently used. I would change this back, but it would take extra effort, and I am lazy.

platforms/sound/openal/SoundStreamAL.cpp
BrentDaMage: How would you suggest we abstract it out?

TheBrokenRail: Maybe make an API where the core code can ask for a platform-agnostic buffer object, then upload the decoded data to that?

But putting decoding code in every sound backend (and we have three!) seems like a bad idea.

BrentDaMage: How does it look now?

BrentDaMage:
Damn, forgot to push. This should be all good to go imo, I'd just like someone to review it before it's merged upstream.

BrentDaMage:
@TheBrokenRail could you please update the README.md file to reflect the new script you added?

TheBrokenRail:
Done!

PR #155, opened 2025-06-04
Monster.cpp implementation

break-core:
Don't let the branch name fool you, I was gonna work on Zombie but instead worked on Monster just so we had a base for future monsters if any are added.

Some important changes:
- `canSpawn()` is no longer `const` because we need to use Random in Monster
- `field_114` in `Mob` is now `m_attackTime` (name confirmed from b1.2_02)
- `checkHurtTarget()` in PathfinderMob has `void` return instead of `bool` (matches with b1.2_02 and works with Monster)
- Monster is pretty much fully implemented

Info provided by b1.2_02/b1.7.3

### DIFF
--- a/source/world/entity/Monster.cpp
+++ b/source/world/entity/Monster.cpp
@@ -1,0 +1,78 @@
+#include "world/entity/Monster.hpp"
+
+Monster::Monster(Level* pLevel) : PathfinderMob(pLevel)
+{
+	m_health = 20;
+}
+
+void Monster::aiStep()
+{
+	float brightness = getBrightness(1.0f);
+
+	if (brightness > 0.5f) 
+	{
+		m_noActionTime += 2;
+	}
+
+	PathfinderMob::aiStep();
+}
+
+void Monster::tick()
+{
+	PathfinderMob::tick();
+
+	// Remove any monsters when in peaceful mode
+	if (m_pLevel->m_difficulty == 0) {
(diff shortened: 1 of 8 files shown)


source/world/entity/Monster.hpp
```diff
@@ -0,0 +1,19 @@
...
+	void aiStep() override;
+	void tick() override;
+	Entity* findAttackTarget() override;
+	bool hurt(Entity* pCulprit, int damage) override;
+	virtual void checkHurtTarget(Entity*, float) override;
+	virtual float getWalkTargetValue(const TilePos& pos) const override;
+	bool canSpawn() override;
+
+protected:
+	int m_attackDamage = 2;
```
BrentDaMage: `m_attackDamage` needs to be initialized in the constructor to maintain C++03 compatibility.

source/world/entity/Monster.cpp
```diff
@@ -0,0 +1,85 @@
...
+}
+
+float Monster::getWalkTargetValue(const TilePos& pos) const
+{
+	float brightness = m_pLevel->getBrightness(pos);
+
+	return 0.5f - brightness;
+}
+
+bool Monster::canSpawn()
```
BrentDaMage, line 74: Not all code paths return a value.

BrentDaMage (approved):
LGTM

PR #156, opened 2025-06-04
cloud culling fix

known81331:
ditto, fixed cloud overdrawing blocks in distant horizon


BrentDaMage (approved):
Fixed OpenGL and OpenGL ES function call incompatibilities. Bugfixes are tested and confirmed working.

PR #158, opened 2025-06-05
Creeper and CreeperRenderer implementation

break-core:
so yeah, I implemented creeper fully. this is based on b1.2_02 and works pretty much identically. for anyone who wants to play around a bit, there's a `MAX_SWELL` define in Creeper.hpp which lets you control how much a creeper can swell up.


BrentDaMage:
@break-core we need to make it so Monsters do not pathtrace to the player when they're in creative mode

uniformization:
image
image
image
image
image

BrentDaMage:
@uniformization if a creative player was around a group of survival mode players, would this not cause the mob to become blind to the survival players?

TheBrokenRail:
Does ReMCPE even support having players with different game modes in the same world?

uniformization:
> @uniformization if a creative player was around a group of survival mode players, would this not cause the mob to become blind to the survival players?

the mob wont be blind to survival players, rather get distracted on creative players because of an issue where findAttackTarget can return a creative player, let me fix it
edit: fixed

BrentDaMage:
@uniformization could you not just add a bool parameter to Level::getNearestPlayer() to exclude players where Player::isCreative() is true? This way we're doing the exclusion in the search, rather than after.

PR #179, opened 2025-08-29
NBT-Based Level Saving/Loading (Level Storage Version 2)

BrentDaMage:
Added NBT-based saving/loading for players, entities, and levels.
All old worlds will be upgraded from V1 to V2 automatically when loaded in-game.

### DIFF
--- a/GameMods.hpp
+++ b/GameMods.hpp
@@ -20,22 +20,23 @@
 // Features (major changes)
 
 // Enhancements (minor changes)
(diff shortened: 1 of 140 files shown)


BrentDaMage:
I have yet to test this with actual 0.2.1 worlds.

BrentDaMage:
Tested and confirmed working with 0.2.1 worlds.

PR #182, opened 2025-09-07
Fix OpenGL Linking With CMake

TheBrokenRail:
image

### DIFF
--- a/platforms/sdl/CMakeLists.txt
+++ b/platforms/sdl/CMakeLists.txt
@@ -60,7 +60,7 @@ if(ANDROID)
     target_compile_definitions(reminecraftpe-core PUBLIC USE_GLES)
 else()
     find_package(OpenGL REQUIRED)
-    target_link_libraries(reminecraftpe-core PUBLIC ${OPENGL_gl_LIBRARY})
+    target_link_libraries(reminecraftpe-core PUBLIC OpenGL::GL)
     if(EMSCRIPTEN)
         # Use WebGL 2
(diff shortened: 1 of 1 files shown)


BrentDaMage:
There's a reason I changed this the other day, it's because doing OpenGL::GL breaks building via CMake on Mac OS 10.8 thru Mac Ports.

TheBrokenRail:
Doesn't the macOS build not even use CMake? Anyway, you're really supposed to use the `OpenGL::*` targets now:

image

Anyway, between breaking building on 13-year-old version of macOS and breaking building on modern Linux, I would prefer breaking the ancient macOS.

BrentDaMage:
The real question is, why did they break building on modern Linux? I thought CMake maintained backwards-compatibility. What solution do they offer to this issue in breaking it?

TheBrokenRail:
OK, so CMake's behavior is controlled by [`CMP0072`](https://cmake.org/cmake/help/latest/policy/CMP0072.html). Because we have our minimum version of CMake set to `3.16.0` (released in 2019), this policy is set to `NEW`.

This means:
- If `libOpenGL` (the GLVND-based version) is available, it will be used and `${OPENGL_gl_LIBRARY}` will be blank.
- Otherwise, legacy `libGL` will be used and `${OPENGL_gl_LIBRARY}` will be set.

I'm not sure why updating to Ubuntu 25.04 made CMake realize I had GLVND installed, but this is the correct behavior.

BrentDaMage:
Works for me.

PR #184, opened 2025-09-07
fix text input

known81331:
its been broken since the merge of del & backspace from keypressed to charpressed of text control

what was broken:
- pressing Y = pressing delete
- pressing backspace = pressing backspace + delete

how it has been fixed:
- use ansi delete code instead of sdl's
- added missing breaks


BrentDaMage:
I forgot to even finish this code. Thanks for catching it, the solution works great.

PR #193, opened 2025-09-29
Fixed networking

BrentDaMage:
Should fix endian swap for networking.


BrentDaMage:
also removed SDL-specific implementations for file read and writes, because it was causing memory issues or some shit, and is just bonus maintenance

TheBrokenRail:
The SDL-specific file I/O code is needed for Android.

BrentDaMage:
> The SDL-specific file I/O code is needed for Android.

Why?

BrentDaMage:
Also, I'm fairly certain it leaks memory, unless the std::string being passed out automatically deallocates the buf variable from the heap.

TheBrokenRail:
> > The SDL-specific file I/O code is needed for Android.
> 
> Why?

It automatically handles retrieving data from the Android asset repository.

`getPatchData` however, can be removed. I don't know why an SDL-specific version was ever added, but it is not needed. `AppPlatform::getPatchData` is perfectly fine (and correctly handles freeing memory).

BrentDaMage:
> Also, I'm fairly certain it leaks memory, unless the std::string being passed out automatically deallocates the buf variable from the heap.

Sorry, was thinking of the getPatchData function. I believe the readAssetFile one should be fine. Re-adding it...

PR #197, opened 2025-10-17
Network Protocol V3 Support

### LINKED ISSUE #144: Network Protocol V3 Support
Network protocol V3 support is required to replicate entities.

BrentDaMage:
closes #144

### DIFF
--- a/source/world/item/DoorItem.cpp
+++ b/source/world/item/DoorItem.cpp
@@ -18,7 +18,7 @@ DoorItem::DoorItem(int id, Material* pMtl) : Item(id)
 	m_pMaterial = pMtl;
 }
 
-bool DoorItem::useOn(ItemInstance* inst, Player* player, Level* level, const TilePos& pos, Facing::Name face)
+bool DoorItem::useOn(ItemInstance* inst, Player* player, Level* level, const TilePos& pos, Facing::Name face) const
 {
 	if (face != Facing::UP)
 		return false;
--- a/source/world/item/DoorItem.hpp
+++ b/source/world/item/DoorItem.hpp
@@ -14,7 +14,7 @@ class DoorItem : public Item
 public:
 	DoorItem(int id, Material* pMtl);
 
-	virtual bool useOn(ItemInstance*, Player*, Level*, const TilePos& pos, Facing::Name face);
+	bool useOn(ItemInstance*, Player*, Level*, const TilePos& pos, Facing::Name face) const override;
 
 public:
 	Material* m_pMaterial;
--- a/source/world/item/Inventory.cpp
+++ b/source/world/item/Inventory.cpp
@@ -185,7 +185,7 @@ void Inventory::prepareSurvivalInventory()
 	addCreativeItem(TILE_MUSHROOM_1);
 	addCreativeItem(TILE_MUSHROOM_2);
 	addCreativeItem(TILE_CACTUS);
-	addCreativeItem(TILE_REEDS);
+	addCreativeItem(ITEM_REEDS);
 
 #ifdef MOD_POCKET_SURVIVAL
 	for (int i = 0; i < getNumSlots(); i++)
--- a/source/world/item/TilePlanterItem.cpp
+++ b/source/world/item/TilePlanterItem.cpp
@@ -15,7 +15,7 @@ TilePlanterItem::TilePlanterItem(int id, int place) : Item(id)
 	m_tile = Tile::tiles[place]->m_ID;
 }
 
-bool TilePlanterItem::useOn(ItemInstance* instance, Player* player, Level* level, const TilePos& pos, Facing::Name face)
+bool TilePlanterItem::useOn(ItemInstance* instance, Player* player, Level* level, const TilePos& pos, Facing::Name face) const
 {
 	TilePos tp(pos);
 
--- a/source/world/item/TilePlanterItem.hpp
+++ b/source/world/item/TilePlanterItem.hpp
@@ -14,7 +14,7 @@ class TilePlanterItem : public Item
 public:
 	TilePlanterItem(int id, int place);
 
-	virtual bool useOn(ItemInstance*, Player*, Level*, const TilePos& pos, Facing::Name face);
+	bool useOn(ItemInstance*, Player*, Level*, const TilePos& pos, Facing::Name face) const override;
 
 public:
 	int m_tile;


source/client/gui/screens/SelectWorldScreen.cpp
```diff
@@ -134,7 +134,12 @@ void SelectWorldScreen::tick()
 			}
 		}
 
-		m_pMinecraft->selectLevel(levelUniqueName, levelNickname, seed);
+		GameType gameType = GAME_TYPE_CREATIVE;
+		if (userInput.size() > 2 && userInput[2].compare("survival") == 0)
```
TheBrokenRail, line 138: Is this even necessary? IIRC `LevelSettings` is ignored if the world already exists, and if this is in the `SelectWorldScreen`, it definitely already exists.

source/client/renderer/entity/HumanoidMobRenderer.cpp
```diff
@@ -64,8 +64,10 @@ void HumanoidMobRenderer::render(Entity* pEntity, const Vec3& pos, float f1, flo
 		ItemInstance* item = player->getSelectedItem();
 		m_pHumanoidModel->m_bHoldingRightHand = item != nullptr;
 	}
+
 	if (pEntity->isSneaking())
 	{
+		m_pHumanoidModel->m_bSneaking = true;
```
TheBrokenRail, line 70: This should be set to `false` after rendering, otherwise it can cause a small bug:

1. Switch to third-person
2. Start sneaking
3. Switch to first-person
4. Stop sneaking
5. Notice that the first-person hand is rendered weirdly.

This is because `m_pHumanoidModel->m_bSneaking` was set to `true` while in third-person, but was never reset (because `render()` was never called in first-person).

And yes, having "whether the model is sneaking" being effectively global state is horrifying.

I know about this bug because I had to patch it out of MCPI-Reborn.

source/network/NetEventCallback.hpp
```diff
@@ -9,53 +9,54 @@
...
-	virtual void handle(const RakNet::RakNetGUID&, RemoveEntityPacket*);
-	virtual void handle(const RakNet::RakNetGUID&, MovePlayerPacket*);
-	virtual void handle(const RakNet::RakNetGUID&, PlaceBlockPacket*);
-	virtual void handle(const RakNet::RakNetGUID&, RemoveBlockPacket*);
-	virtual void handle(const RakNet::RakNetGUID&, UpdateBlockPacket*);
-	virtual void handle(const RakNet::RakNetGUID&, RequestChunkPacket*);
-	virtual void handle(const RakNet::RakNetGUID&, ChunkDataPacket*);
-	virtual void handle(const RakNet::RakNetGUID&, PlayerEquipmentPacket*);
-	virtual void handle(const RakNet::RakNetGUID&, LevelDataPacket*);
+	virtual void handle(const RakNet::RakNetGUID&, LoginPacket*) {}
```
TheBrokenRail, line 31: Why not use a pure virtual function (`virtual void handle(...) = 0;`)?

source/world/entity/Entity.cpp
```diff
@@ -71,11 +70,31 @@ Entity::Entity(Level* pLevel)
...
+	m_entityData.clear();
+}
+
+bool Entity::getSharedFlag(SharedFlag flag) const
+{
+	return (getEntityData().get<int8_t>(DATA_SHARED_FLAGS_ID) & 1 << flag) != 0;
+}
+
+void Entity::setSharedFlag(SharedFlag flag, bool value)
+{
```
TheBrokenRail, line 87: In real MCPE, `setSharedFlag` is part of `Mob`. Being part of `Entity` is undoubtedly better though.

BrentDaMage: In Java, `setSharedFlag` is a part of `Entity`.

source/client/gui/screens/SelectWorldScreen.cpp
BrentDaMage: It's not necessary, since the entire function is unused. I just decompiled it from 0.2.1 for fun.

source/network/NetEventCallback.hpp
BrentDaMage: Because not all packets are handled by `ClientSideNetworkHandler` or `ServerSideNetworkHandler` respectively. For example, `ClientSideNetworkHandler` shouldn't handle `LoginPacket`, and doesn't have an implementation for it. If these were pure virtual functions, the network handlers couldn't be compiled.

source/client/gui/screens/SelectWorldScreen.cpp
BrentDaMage: It's a leftover from when Mojang used native modal GUIs for world creation on the world selection screen.

source/client/renderer/entity/HumanoidMobRenderer.cpp
BrentDaMage: Fixed.

BrentDaMage:
@iProgramMC @TheBrokenRail I'm thinking of increasing the version number of ReMCPE from 0.1.0 to 0.2.0 to reflect the default network protocol support. I feel like not being able to connect to actual 0.1.0 clients could be confusing to users. What are your thoughts?

TheBrokenRail:
That sounds reasonable to me.

But I'm also a little confused. There seems to be a lot of code meant to support multiple protocols, like:
```cpp
	if (m_serverProtocolVersion >= 3)
	{
		ReadyPacket* pReadyPkt = new ReadyPacket(1);
		m_pRakNetInstance->send(pReadyPkt);
	}
```

Or:
```cpp
#if NETWORK_PROTOCOL_VERSION >= 3
	char pitch, yaw;
	bs.Read(pitch);
	bs.Read(yaw);
	bs.Read(m_itemId);
	bs.Read(m_itemAuxValue);
	m_rot.x = PacketUtil::Rot_charToDegrees(pitch);
	m_rot.y = PacketUtil::Rot_charToDegrees(yaw);
#endif
```

If supporting both protocols is not the goal, I'd suggest removing all the old code to simplify it. And if it is the goal, then it's not really a breaking charge.

Just my two cents.

BrentDaMage:
@TheBrokenRail To clarify, the old network protocol versions are still supported when the `NETWORK_PROTOCOL_VERSION` macro is adjusted. However, because Mojang decided to completely change all but one of the packet IDs between V2 and V3, support for multiple protocol versions on runtime has not been added.

PR #205, opened 2025-11-28
Fix entity's chunk Y-coordinate not updating

### LINKED ISSUE #172: Placing blocks while being in certain positions of the world doesn't work correctly
https://youtu.be/5LsN_iRmz70

bluepilledgreat:
Fixes #172

### DIFF
--- a/source/world/level/levelgen/chunk/LevelChunk.cpp
+++ b/source/world/level/levelgen/chunk/LevelChunk.cpp
@@ -324,18 +324,21 @@ void LevelChunk::updateEntity(Entity* pEnt)
 		return;
 	}
 
-	std::vector<Entity*>::iterator it = std::find(m_entities[oldYCoord].begin(), m_entities[oldYCoord].end(), pEnt);
-	if (it != m_entities[oldYCoord].end())
+	std::vector<Entity*>& oldTerrainLayer = m_entities[oldYCoord];
+	std::vector<Entity*>& newTerrainLayer = m_entities[newYCoord];
+
+	std::vector<Entity*>::iterator it = std::find(oldTerrainLayer.begin(), oldTerrainLayer.end(), pEnt);
+	if (it != oldTerrainLayer.end())
 	{
-		m_entities[oldYCoord].erase(it);
+		oldTerrainLayer.erase(it);
 	}
 	else
 	{
 		assert(false);
 	}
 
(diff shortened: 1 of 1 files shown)


Vimd0:
> Fixes #172

OMG thank you! This has been a bug for the longest time 😭

source/world/level/Level.cpp
```diff
@@ -1558,15 +1558,20 @@ void Level::tick(Entity* pEnt, bool b)
 
 		if (hasChunk(cp))
 		{
-			pEnt->m_bInAChunk = true;
```
BrentDaMage, line 1561: Why remove this?

source/world/level/levelgen/chunk/LevelChunk.cpp
```diff
@@ -302,6 +302,43 @@ void LevelChunk::addEntity(Entity* pEnt)
...
+	}
+
+	std::vector<Entity*>::iterator it = std::find(m_entities[oldYCoord].begin(), m_entities[oldYCoord].end(), pEnt);
+	if (it != m_entities[oldYCoord].end())
+	{
+		m_entities[oldYCoord].erase(it);
+	}
+	else
+	{
+		assert(false);
```
BrentDaMage, line 337: Did you forget to return here?

source/world/level/levelgen/chunk/LevelChunk.cpp
```diff
@@ -302,6 +302,43 @@ void LevelChunk::addEntity(Entity* pEnt)
...
+		return;
+	}
+
+	if (oldYCoord < 0 || oldYCoord > 7)
+	{
+		assert(false);
+		return;
+	}
+
+	std::vector<Entity*>::iterator it = std::find(m_entities[oldYCoord].begin(), m_entities[oldYCoord].end(), pEnt);
```
BrentDaMage: Improve readability by making local variables for `m_entities[oldYCoord]` and `m_entities[newYCoord]` std::vector references. Perhaps call them `newTerrainLayer` and `oldTerrainLayer`?

source/world/level/Level.cpp
bluepilledgreat: `addEntity` sets it to true anyways, so there is no real point in doing it here

source/world/level/levelgen/chunk/LevelChunk.cpp
bluepilledgreat: Did not forget, it not being in the original vector is not intended but it shouldn't mean we should just entirely stop as it should be in one of the vectors

PR #207, opened 2025-11-30
Fix bad ToChunkCoordinate

bluepilledgreat:
Sorry

Fixes chunks with ChunkPos X/Z=0 duplicating


er2off:
Checked this on godbolt, it's 1-2 instructions shorter. Don't know if anyone really notice performance difference but OK.

bluepilledgreat:
> Checked this on godbolt, it's 1-2 instructions shorter. Don't know if anyone really notice performance difference but OK.

This PR isnt necessarily for performance. This fixes a bug with chunks duplicating that I accidentally introduced with #205 while cleaning up some of the code.

PR #209, opened 2025-12-05
Rendering Abstraction Layer (HAL Renderer)

### LINKED ISSUE #137: Rendering Abstraction Layer (HAL Renderer)
We need a rendering abstraction layer that will replace OpenGL's direct use in ReMCPE with an underlying implementation in OpenGL (for 99% of hardware), Direct3D (for Microsoft hardware), GX (for wahoo-type hardware), etc.
Mojang's HAL (Hardware Abstraction Layer) for rendering seems to be the best solution for this. A decompilation of the HAL from MCPE 0.12.1 is currently in the works [here](https://github.com/BrentDaMage/mcpe-engine/tree/main/src-deps/Renderer/HAL).

Please reach out to me if you need the IDA DB files for the binary, as I've been documenting tons of functions and structures in there.

BrentDaMage:
Closes #137

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,9 +46,8 @@ else()
     link_libraries(Threads::Threads)
 endif()
 
-
-# Disable thread local storage for rapidjson (needed for C++98 & iOS)
-add_compile_options(-DRAPIDJSON_NO_THREAD_LOCAL)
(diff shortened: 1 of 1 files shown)


BrentDaMage:
Will test Mac OS X support tomorrow, since it's what errored when reloading levels.

PR #212, opened 2025-12-06
Always use C++98

Un1q32:


TheBrokenRail:
Aren't we sticking with with C++03? Because C++98 is *really* old.

Un1q32:
Isn't 03 just 98 with a few addendums?  In clang `-std=c++98` is an alias for c++03.

Un1q32:
cmake actually doesnt even let you set the std to 03 lol, probably cuz they're basically the same and -std=c++98 is more portable (gcc 4.2 doesnt accept 03)

PR #214, opened 2025-12-06
Relicense to AGPLv3

Un1q32:
We support WASM builds that run in the browser, and there are plans to add our own non-Mojang assets, this means it's likely inevitable that people are going to make their own instances. Many projects have had issues with other people making forks that just rebrand the project, and hosting said forks to steal attention from the upstream project, an AGPL license would offer some deterrence without making the project no longer open source.


TheBrokenRail:
Personally, I prefer more permissive licenses, like MIT and what ReMCPE has right now. I like the idea that anyone can do whatever they want with my code. They are also just way simpler.

Yeah, that means someone could make a competing closed-source fork if they wanted to. But that also wouldn't affect me at all. All the power to them. (Also, if someone actually tried to make money off a closed-source fork, Mojang would almost certainly DMCA them.)

Of course, I cannot speak for other ReMCPE contributors. This is just my personal opinion.

Un1q32:
The way I see it is why would anyone ever want to make a closed source fork except for the purpose of taking credit for something that isn't theirs. I guess someone might want to make their own totally different minecraft-like game using this as the engine but I find that unlikely. And even in that case there's no real reason they couldn't just open source it unless they plan to make money off it, which as you pointed out would likely get them sued.

BrentDaMage:
I'm pretty indifferent here, but I do have some things that are worth considering:

Personally, I would like to see this project and its derivatives always remain open-source, so they can be maintained by the community long into the future. Keeping the license as it is could allow a closed-source version of the codebase to become the predominant one in the future, which would cripple community control.
However, if someone were to create a fork of the project that uses a proprietary/private SDK, this would prevent them from distributing builds to the public, which we wouldn't want to stand in the way of.

Honestly, I think a closed-source version of the project becoming the predominant one is highly unlikely and would be prevented naturally, so I'm fine with keeping the license as-is to give people the most freedoms.

BrentDaMage:
@Vimd0 @EGAMatsu @Wilyicaro @uniformization @Stommm @MFDGaming 
Would like to know your thoughts on changing licenses, since you have all contributed a substantial amount to the project.

Vimd0:
> @Vimd0 @EGAMatsu @Wilyicaro @uniformization @Stommm @MFDGaming Would like to know your thoughts on changing licenses, since you have all contributed a substantial amount to the project.

I'm fine with this

MFDGaming:
> @Vimd0 @EGAMatsu @Wilyicaro @uniformization @Stommm @MFDGaming Would like to know your thoughts on changing licenses, since you have all contributed a substantial amount to the project.

I'm not really fond of GPL, but whatever, I personally don't mind

Un1q32:
> if someone were to create a fork of the project that uses a proprietary/private SDK, this would prevent them from distributing builds to the public, which we wouldn't want to stand in the way of.

Actually, I believe this wouldn't apply, see the "System Libraries" section.
(Not a lawyer not legal advice)

Un1q32:
Also, said projects could just distribute only source code and require the user to compile the binary themselves, providing an easy script that does so automatically.  We could also just turn a blind eye to anyone distributing binaries in this particular case.

Wilyicaro:
> @Vimd0 @EGAMatsu @Wilyicaro @uniformization @Stommm @MFDGaming Would like to know your thoughts on changing licenses, since you have all contributed a substantial amount to the project.

I'm okay with that

Un1q32:
Closing because apparently this would affect mods.

iProgramMC:
I initially chose a derivative of the BSD license with one clause because I felt that Mojang should abide by our license. When I chose the license this was nothing more than a decomp of MCPE v0.1.0, so I believed it made sense at the time. I'd relicense using the MIT license, but I would need the permission of every contributor.

TheBrokenRail:
I'm not a lawyer, but I think you only need permission from everyone when moving to a more permissive license.

Because changing BSD to MIT only adds restrictions. MIT doesn't allow anything that would violate the BSD 0-clause. It would just add restrictions to any new contributions (because old contributions would still technically be licensed under BSD).

Meanwhile, changing GPL to MIT would require permission. Because MIT allows a lot of things that the GPL does not. To relicense it, you would have to *remove* restrictions from old contributions, which involves contacting everyone and asking permission.

Stommm:
> @Vimd0 @EGAMatsu @Wilyicaro @uniformization @Stommm @MFDGaming Would like to know your thoughts on changing licenses, since you have all contributed a substantial amount to the project.

I'm OK with whatever licensing changes will happen 👍

iProgramMC:
> I'm not a lawyer

Me neither.

> Because changing BSD to MIT only adds restrictions.

From reading the text I guess the BSD 1-clause license we use and MIT have roughly the same terms: you may distribute in source and binary form, but you must keep the license notice in source distributions, and also we don't provide warranty over the content of this repo.

But our license doesn't specifically state that you are allowed to sell and sublicense the project, and it doesn't specifically state that you must provide the contents of the license in binary distributions either.

PR #215, opened 2025-12-06
fix compiler warns about abs

Un1q32:
For iOS builds Clang warns about using the integer abs function with float args, according to iProgramCPP this is supposed to do some C++ shit and call a version of abs that uses float args and return value, but according to Clang it isn't so we can just use the C fabsf function instead.

### DIFF
--- a/source/client/player/input/UnifiedTurnBuild.cpp
+++ b/source/client/player/input/UnifiedTurnBuild.cpp
@@ -10,7 +10,6 @@
 #include "Multitouch.hpp"
 
 #include "world/entity/Player.hpp"
-#include "common/Utils.hpp"
 
 UnifiedTurnBuild::UnifiedTurnBuild(int a, int width, int height, float d, float e, IInputHolder* pHolder) :
 	field_C(a),


source/common/Utils.hpp
```diff
@@ -556,3 +556,9 @@ bool DeleteDirectory(const std::string& name, bool unused);
 uint8_t* ZlibInflateToMemory(uint8_t* pInput, size_t compressedSize, size_t decompressedSize);
 uint8_t* ZlibDeflateToMemory(uint8_t* pInput, size_t sizeBytes, size_t *compressedSizeOut);
 uint8_t* ZlibDeflateToMemoryLvl(uint8_t* pInput, size_t sizeBytes, size_t* compressedSizeOut, int level);
+
+static inline float abs_float(float arg) {
```
bluepilledgreat: Why not use fabs?

bluepilledgreat (changes requested):

source/common/Utils.hpp
Un1q32: Was worried about name conflicts and didnt feel like grepping the source

bluepilledgreat: > Was worried about name conflicts and didnt feel like grepping the source

I meant the built-in `fabs` function. I don't know why you're adding your own function.

Un1q32: Becuase I didn't know that was a thing until just now lol

bluepilledgreat (approved):

PR #222, opened 2025-12-07
Improvements to the iOS build

Un1q32:
- Enables LTO when possible on the cmake build.
- Makes the Xcode build use codesign instead of ldid because codesign is always available on macOS, but ldid might not be.
- Makes the cmake build script use codesign on macOS instead of ldid to save having to build it there.
- Fixes issues with the janky nproc function thing in the cmake build script.
- Changes the .app name in the cmake script from ReMCPE.app to minecraftpe.app to match the Xcode build.
- Changes how icon.png is added, not renaming it anymore and avoiding duplicating it unnecessarily.
- Stops renaming default.png to default8.png, since we automatically fallback it's no longer needed.
- Raises minimum version to 3.1 since 3.0 doesn't work.
- Removes armv7s from Xcode build to speed up builds and reduce binary size.
- Adds $DEBUG variable for doing debug builds with the cmake build script.

### DIFF
--- a/platforms/ios/build.sh
+++ b/platforms/ios/build.sh
@@ -1,4 +1,5 @@
 #!/bin/sh
+# shellcheck disable=2086
 set -e
 
 [ "${0%/*}" = "$0" ] && scriptroot="." || scriptroot="${0%/*}"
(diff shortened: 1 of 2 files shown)


platforms/ios/build.sh
```diff
@@ -112,7 +111,7 @@ for target in $targets; do
     cd build
 
     cmake .. \
-        -DCMAKE_BUILD_TYPE=Release \
+        -DCMAKE_BUILD_TYPE=MinSizeRel \
```
BrentDaMage: Don't we not need this anymore?

Un1q32: I kept it cuz Xcode builds with -Os by default and I wanna be close to Xcode, but if you think we should do -O3 I can change it back.

BrentDaMage: Make it more optimized as long as it doesn't cause problems. I'm sure it will help with performance.

PR #224, opened 2025-12-08
Fix issues with TARGET_OS_* macros

Un1q32:
Apple clang doesn't define the TARGET_OS_* macros, I was getting warns when using the cmake build script on a mac, these macros should work for any compiler targetting Darwin, including GCC 4.


BrentDaMage (approved):
LGTM

PR #232, opened 2025-12-13
Cache LevelChunk in Chunk::rebuild

bluepilledgreat:


TheBrokenRail:
Wouldn't this cause issues if a `Chunk` consists of multiple `LevelChunk`s?

bluepilledgreat:
> Wouldn't this cause issues if a `Chunk` consists of multiple `LevelChunk`s?

No, the region only encompasses one LevelChunk

source/client/renderer/Chunk.cpp
```diff
@@ -155,7 +156,7 @@ void Chunk::rebuild()
 			{
 				for (tp.x = min.x; tp.x < max.x; tp.x++)
 				{
-					TileID tile = region.getTile(tp);
```
BrentDaMage, line 158: Please comment out line 158 instead of removing it, then put comment above explaining why this change was made. In the event of a regression way down the line, it will make this easier to find.

PR #235, opened 2025-12-18
Fix lag caused by HAL shader enablement

rep-stosd:
The root cause of mysterious lag that brings every mobile and computer to its knees is the misalignment of vertex attribute.

It's related with the deprecation of Texture_1 in shaders, as that's used for modern ambient & block light ramps (LUT) and we don't need it here since we prefer monochromatic lights present in versions <= 0.9.5, so we disable Texture_1 from game.
While the intent of this act is good (to save ram and OpenGL transactions), removing its UV attribute unfortunately broke the harmony of align by 16 the original vertex attrbiute had.
Unfortunately GLSL doesn't offer a way to directly skip offsets in its vertex attributes like HLSL would, so a 4th attribute needs to bee there to keep it aligned by 16 either way, so we keep Texture_1 UV even if we don't actually use it.

Additionally, the PR has added/modified the following to the HAL:
- #include support
- uniforms introduced in later versions of HAL (partially)
- enforce GLES 2.0 on iOS

I apologize for not keeping commits organized, but cherrypick anything as you please. Only the last 2 commits are required for the main fix.

### DIFF
--- a/source/renderer/platform/ogl/ShaderPrecision.cpp
+++ b/source/renderer/platform/ogl/ShaderPrecision.cpp
@@ -91,7 +91,6 @@ void Precision::BuildHeader(std::ostringstream& stream)
     stream << "#define POS3 " << AtLeast(23) << " vec3\n";
     if (gl::isOpenGLES())
         stream << "precision "    << AtLeast(9)  << " float;\n";
-        
 }
 
 #endif // FEATURE_GFX_SHADERS


source/renderer/hal/ogl/ShaderProgramOGL.cpp
```diff
@@ -43,6 +43,11 @@ void ShaderProgramOGL::deleteShader()
 
 std::string _getVersionMacro()
 {
+// for compatibility reason
+#if defined(__APPLE__) && TARGET_OS_IPHONE
+    return "\n";
```
BrentDaMage, line 48: Why would this not be needed?

game/assets/shaders/glsl/renderchunk.vertex
```diff
@@ -12,7 +12,7 @@ uniform float FAR_CHUNKS_DISTANCE;
 attribute POS4 POSITION;
 attribute vec4 COLOR;
 attribute vec2 TEXCOORD_0;
-//attribute vec2 TEXCOORD_1;
+attribute vec2 TEXCOORD_1;
```
BrentDaMage: I guess this isn't needed for the fragment shader? Also, I'm having a really hard time understanding why on earth we would be required to have a shader require a specific amount of attributes? Is there no way to handle the alignment properly, no matter the shader's attributes?

source/renderer/hal/ogl/ShaderProgramOGL.cpp
rep-stosd: sorry there was a confusion made when i said "enforce GLES 2.0" where i actually meant "enforce GLSL that opengl es 2.0 likes"

the multiversion tag when resolved to `#version 110` conflicts with `precision` settings that the parser later sets, leading to compiler error that would either complain about undefined precision with it removed or complain about undefined `float` et. al when kept.

`#version 100` is not an option, unfortunately

game/assets/shaders/glsl/renderchunk.vertex
rep-stosd: whether this is needed for the fragment shader is up to your own discretion depending on how you wish the aesthetics to look like. its currently not practically needed since we're concerning b1.7 specifically.

however it needs to be kept to ensure the attribute list is a multiple of 16 so memory stay alined in gpu. you can imagine the entire attribute list as the vs_in struct in d3d11/hlsl, which its size needs to be maintained as required. 

simple math shows that 
sizeof(vec4+vec4+vec2+vec2) = 16+16+8+8 = 48.
sizeof(vec4+vec4+vec2) = 40 and 40 is not divisible by 16 hence why we keep it.

the name of TEXCOORD_1 may be misleading since there practically isn't a use for it, in reality it can be named to` __padd0` or whatever necessary, but it still needs to be referenced in the shader at least once (so it doesn't get culled which defeats the purpose) and be enabled on cpu's vertex buffer/input layout. since there exists code to already do it for tex1 i figure its easier to keep it as is than to maintain a separate alignment helper that works with `__padd0` in theory.

rep-stosd: on newer versions of opengl (3.0+) GLSL supports descriptors such as `layout (location = 0) in vec3 ….` which explicitly mandates the size of the `vs_in` struct philosophy and therefore solving the issue.

this isn't applicable for us since we're targeting opengl 2.x hence workarounds (pronounced HACKS) are required

BrentDaMage: Wouldn't the entity vertex shader have the exact same issue?

rep-stosd: yes, however entities are not yet significant in the game so impact isn't as large as renderchunk has. the original mcpe shader solves this issue by making `POSITION` mediump which makes it 8 bit to fit with rest of attribute list.

we can add `TEXCOORD_1` to our shader to fix the problem if required

BrentDaMage: Please put a comment in entity.vertex and renderchunk.vertex indicating these issues. As far as entity.vertex, I'm really not a fan of wasting more memory on another texcoord attribute, but we also don't have precision available to us on earlier non-GLES implementations. I'm assuming there's really no good solution that doesn't waste memory?

game/assets/shaders/glsl/entity.vertex
```diff
@@ -14,6 +14,7 @@ uniform vec2 GLINT_UV_SCALE;
 
 attribute POS4 POSITION;
 attribute vec2 TEXCOORD_0;
+attribute vec2 TEXCOORD_1; // attribute padding (see PR #235)
```
BrentDaMage, line 17: We can't just add this without calling `Tesselator::tex1()`, otherwise `glDrawElements()` will crash.

BrentDaMage (changes requested):

game/assets/shaders/glsl/renderchunk.vertex
rep-stosd: no real solution before at least d3d11, sorry! :/

with parallelism it comes alignment, it is unavoidable, this applies to cuda/openmp/compute shader as well. putting gaps in vertex buffer is the most straightforward way to pad data.
see: https://www.songho.ca/misc/alignment/dataalign.html

game/assets/shaders/glsl/entity.vertex
rep-stosd: currently afk, pls do me a favor

Un1q32:
Tested on an iPhone 5 running 10.3.4, this greatly improves performance with shaders enabled, seems to be roughly on par with no shaders.
When enabling shaders it crashes on launch for older stuff, like my iPod Touch 4 on 6.1.6 (this was true before this PR) so shaders shouldn't be forced yet, at least for the 32 bit build.

rep-stosd:
@BrentDaMage should be all good

source/client/renderer/entity/EntityRenderer.cpp
```diff
@@ -111,7 +112,9 @@ void EntityRenderer::render(const AABB& aabb, const Vec3& pos)
 {
 	Tesselator& t = Tesselator::instance;
 	currentShaderColor = Color::WHITE;
+
 	t.begin(24);
+	t.tex1(Vec2::ZERO);
```
BrentDaMage: Correct me if I'm wrong, but Cube or ModelPart do a lot of Tesselator calls, which is what's mainly used when rendering entities, so just putting tex calls here wouldn't really suffice.

BrentDaMage:
Honestly, we should probably just leave the entity stuff alone and work on a better solution. Having to call tex everywhere is kind of aids.

source/client/renderer/entity/EntityRenderer.cpp
rep-stosd: There is no need to have everything call into tex1 as we primarily want the tessellator to recognize the presence of a tex1 regardless of what content is inside the buffer (it can even be uninitialized and that is ok):
`		m_vertexFormat.enableField((mce::VertexField)(mce::VERTEX_FIELD_UV0 + count));`
is what we particularly need

As tested in game I didn't encounter any crash from entity rendering (or with this code removed actually) so I think it is ok we keep it this way

rep-stosd: Replying to your other comment, I would withstand this compromise since 1) there typically isn't too much mobs in game as of right now and 2) the game can only sustain ~700 mobs before experiencing framerate drops. The impact caused by this shader in particular is rather small unlike chunks which bear thousands of vertices each with a dozen of them on screen each frame

rep-stosd: Screen Shot 2025-12-19 at 9 52 01 AM
Screen Shot 2025-12-19 at 9 52 21 AM

there aren't real differences between forward and fixed pipeline in concentrated abundances. if they are scattered then culling will do its job. The difference between keeping tex1 and removing tex1 is also negligible.

Un1q32:
Tested with an iPhone 4S on max settings, when the FPS settled shaders were 1-2 FPS slower on average (was getting mid 40s), so basically the same.

CMakeLists.txt
```diff
@@ -69,12 +69,6 @@ add_subdirectory(source)
 # Load Platform-Specific Code
 add_subdirectory(platforms)
 
-# Assets
```
BrentDaMage, line 72: Why? Do we just not need this?

rep-stosd: accident commit. was conflicting with my environment so i got rid of it. please revert this

source/renderer/platform/ogl/ShaderPrecision.cpp
```diff
@@ -89,8 +89,10 @@ void Precision::BuildHeader(std::ostringstream& stream)
     stream << "#define MAT4 " << AtLeast(23) << " mat4\n";
     stream << "#define POS4 " << AtLeast(23) << " vec4\n";
     stream << "#define POS3 " << AtLeast(23) << " vec3\n";
-    if (gl::isOpenGLES())
+    if (gl::isOpenGLES()) {
+        stream << "#extension GL_OES_standard_derivatives : enable\n";
```
BrentDaMage: Please comment why this is needed.

Un1q32:
Should drop all changes to ShaderPrecision.cpp, its just adding useless whitespace.

source/renderer/platform/ogl/ShaderPrecision.cpp
rep-stosd: 0.12.1 supports derivative operations by default, but since we are targeting to run the game on everything (maybe toasters) and there exists devices that don't support derivative operations (since in gles 2.0 its an extension than a requirement) and i'm too lazy to check for support it can be removed

BrentDaMage: Just do
```cpp
        const std::string& glExtensions = gl::getOpenGLExtensions();
        bool supportsSDs = glExtensions.find("GL_OES_standard_derivatives") != std::string::npos);
```

rep-stosd: ok

source/renderer/platform/ogl/ShaderPrecision.cpp
```diff
@@ -89,8 +89,15 @@ void Precision::BuildHeader(std::ostringstream& stream)
     stream << "#define MAT4 " << AtLeast(23) << " mat4\n";
     stream << "#define POS4 " << AtLeast(23) << " vec4\n";
     stream << "#define POS3 " << AtLeast(23) << " vec3\n";
-    if (gl::isOpenGLES())
+    if (gl::isOpenGLES()) {
+        const std::string& glExtensions = gl::getOpenGLExtensions();
```
BrentDaMage: Does this have to go above the precision specification? If not, can we move it to `ShaderOGL::BuildHeader()`? We may want to use more extensions in the future, so having them all in one place would be good.

Un1q32:
Useless newline in ShaderPrecision.cpp again (and removed the newline at the end)

PR #236, opened 2025-12-19
fix crash on old iOS versions when using libstdc++

Un1q32:


source/client/app/AppPlatform.cpp
```diff
@@ -323,6 +323,11 @@ std::string AppPlatform::getAssetPath(const std::string& path) const
 
 AssetFile AppPlatform::readAssetFile(const std::string& path, bool quiet) const
 {
+	if (path.empty()) {
+		LOG_E("Empty asset file path!");
```
BrentDaMage: Make it a warn.

BrentDaMage: Fix formatting.

Un1q32: What formatting?  What would you like changed?

source/client/app/AppPlatform.cpp
```diff
@@ -323,6 +323,11 @@ std::string AppPlatform::getAssetPath(const std::string& path) const
 
 AssetFile AppPlatform::readAssetFile(const std::string& path, bool quiet) const
 {
+	if (path.empty()) {
```
BrentDaMage: put curly brace on new line

source/client/app/AppPlatform.cpp
```diff
@@ -323,7 +323,8 @@ std::string AppPlatform::getAssetPath(const std::string& path) const
 
 AssetFile AppPlatform::readAssetFile(const std::string& path, bool quiet) const
 {
-	if (path.empty()) {
+	if (path.empty())
```
iProgramMC, line 326: No it's not "evil".

Un1q32: yes it is

PR #237, opened 2025-12-20
Natural mob spawn

### LINKED ISSUE #263: Natural Mob Spawning

rep-stosd:
Allows mobs to spawn naturally in-game. considered implementation of v0.7.1 and referenced code from PEtoLE


source/world/entity/MobCategory.cpp
```diff
@@ -26,4 +26,16 @@ void MobCategory::initMobCategories()
 	MobCategory::monster.m_pSpawnPositionMaterial       = Material::air;
 	MobCategory::creature.m_pSpawnPositionMaterial      = Material::air;
 	MobCategory::waterCreature.m_pSpawnPositionMaterial = Material::water;
+}
+
+MobCategory& MobCategory::GetCategoryByIndex(int i)
```
BrentDaMage: Make this a `MobCategory::ID` enum.

source/world/entity/MobFactory.cpp
```diff
@@ -35,4 +42,25 @@ Mob* MobFactory::CreateMob(EntityType::ID entityType, Level *level)
...
+
+    creatureList.insert(std::make_pair(EntityType::SHEEP,   12));
+    creatureList.insert(std::make_pair(EntityType::PIG,     10));
+    creatureList.insert(std::make_pair(EntityType::CHICKEN, 10));
+    creatureList.insert(std::make_pair(EntityType::COW,      8));
+	
+    waterCreatureList.insert(std::make_pair(EntityType::SQUID, 10));
+}
+
+std::map<EntityType::ID, int> MobFactory::GetMobListOfCategory(EntityCategories::CategoriesMask category) {
```
BrentDaMage: Should return a const reference. You're copying here.

source/world/entity/MobFactory.cpp
```diff
@@ -35,4 +42,25 @@ Mob* MobFactory::CreateMob(EntityType::ID entityType, Level *level)
     }
 }
 
+void MobFactory::initMobLists() {
+    monsterList.insert(std::make_pair(EntityType::SPIDER,   10));
```
BrentDaMage, line 49: What's the int for? Please put a comment.

source/world/entity/MobSpawner.cpp
```diff
@@ -0,0 +1,211 @@
...
+        if ((isFriendly && !allowFriendly) || (!isFriendly && !allowHostile))
+            continue;
+
+        if (level->getEntityCountOfCategory(mask) <= (category.getMaxInstancesPerChunk() * (int)chunksToPoll.size() / 256))
+        {    
+            for (std::set<ChunkPos>::iterator it = chunksToPoll.begin(); it != chunksToPoll.end(); ++it) 
+            {
+                ChunkPos pos = *it;
+
+                std::map<EntityType::ID, int> spawnList = MobFactory::GetMobListOfCategory(mask);
```
BrentDaMage: Should be a const ref.

source/world/entity/MobSpawner.cpp
```diff
@@ -0,0 +1,211 @@
...
+            continue;
+
+        if ((isFriendly && !allowFriendly) || (!isFriendly && !allowHostile))
+            continue;
+
+        if (level->getEntityCountOfCategory(mask) <= (category.getMaxInstancesPerChunk() * (int)chunksToPoll.size() / 256))
+        {    
+            for (std::set<ChunkPos>::iterator it = chunksToPoll.begin(); it != chunksToPoll.end(); ++it) 
+            {
+                ChunkPos pos = *it;
```
BrentDaMage: Make this a const ref.

source/world/entity/MobSpawner.hpp
```diff
@@ -0,0 +1,29 @@
...
+#include "world/level/levelgen/chunk/LevelChunk.hpp"
+#include "world/level/levelgen/chunk/ChunkSource.hpp"
+#include "world/level/storage/LevelStorageSource.hpp"
+#include "world/level/storage/LevelSource.hpp"
+#include "world/level/storage/LevelData.hpp"
+#include "world/level/path/PathFinder.hpp"
+
+class MobSpawner {
+public:
+    static bool IsSpawnPositionOk(MobCategory *category, Level *level, const TilePos& pos);
```
BrentDaMage: Make all of the pointer arguments references instead.

source/world/level/Level.cpp
```diff
@@ -1939,3 +1942,16 @@ float Level::getSunAngle(float f) const
 {
 	return (float(M_PI) * getTimeOfDay(f)) * 2;
 }
+
+int Level::getEntityCountOfCategory(EntityCategories::CategoriesMask category) const
+{
+	int count = 0;
+	for (std::vector<Entity*>::const_iterator it = m_entities.begin(); it != m_entities.end(); it++)
```
BrentDaMage: This is not efficient. We should be maintaining a count of the entities in each category. The counts should only be updated on entity add or remove. This should be a map.

source/world/entity/MobCategory.hpp
```diff
@@ -24,8 +24,11 @@ class MobCategory
 	const Material* getSpawnPositionMaterial() const { return m_pSpawnPositionMaterial; }
 	bool isFriendly() const { return m_bIsFriendly; }
 
+	// custom addition
+	static MobCategory& GetCategoryByIndex(int i);
+
 private:
-	const EntityCategories& m_baseType;
+	const EntityCategories m_baseType;
```
BrentDaMage, line 29: Why?

source/world/entity/MobSpawner.cpp
```diff
@@ -0,0 +1,211 @@
...
+                        if (IsSpawnPositionOk(&category, level, tp)) 
+                        {
+                            Vec3 pPos(tp.x + 0.5, tp.y, tp.z + 0.5);
+
+                            if (!level->getNearestPlayer(pPos, 24.0f, false)) 
+                            {
+                                Vec3 dPos = pPos - level->getSharedSpawnPos();
+                                if (dPos.lengthSqr() >= 576.0f) 
+                                {
+
```
BrentDaMage: Whitespace.

BrentDaMage (changes requested):

source/world/entity/MobCategory.hpp
rep-stosd: will dangle otherwise. see comments from prev commit

BrentDaMage: Then make it a pointer instead.

rep-stosd: memory will leak. we use EntityCategories for the sole purpose of having its category mask, see top of .cpp. making it either a pointer or a reference requires keeping it outside (contrary to making it one-liner in the ctor itself) which brings more hassle than it keeps, since we have to keep a copy either way.

BrentDaMage: Whatever.

source/world/entity/MobSpawner.hpp
BrentDaMage: We are doing this every tick for every mob category. Definitely gonna have a performance impact.

source/world/level/Level.cpp
rep-stosd: image
image

fwiw mcpe has shit code but good point

source/world/entity/MobCategory.cpp
```diff
@@ -28,13 +28,13 @@ void MobCategory::initMobCategories()
 	MobCategory::waterCreature.m_pSpawnPositionMaterial = Material::water;
 }
 
-MobCategory& MobCategory::GetCategoryByIndex(int i) 
+MobCategory& MobCategory::GetCategoryByIndex(MobCategory::ID i)
```
BrentDaMage: Make the reference const.

source/world/entity/MobSpawner.cpp
```diff
@@ -5,14 +5,14 @@
...
 {
     if (!allowHostile && !allowFriendly)
         return;
 
     chunksToPoll.clear();
 
-    for (std::vector<Player*>::const_iterator it = level->m_players.begin(); it != level->m_players.end(); ++it)
+    for (std::vector<Player*>::const_iterator it = level.m_players.begin(); it != level.m_players.end(); ++it)
     {
         Player* player = *it;
```
BrentDaMage: Make const ref.

source/world/entity/MobSpawner.cpp
```diff
@@ -31,42 +31,40 @@ void MobSpawner::tick(Level *level, bool allowHostile, bool allowFriendly)
 
     for (int i = 0; i < MobCategory::numValues; i++)
```
BrentDaMage: You don't even have to really use numValues, you can also just have a `COUNT` entry in the ID enum.

BrentDaMage: Honestly, we shouldn't even have the `GetCategoryByIndex` function, we should just do what we do for EntityTypes where we have them in an array called "all" or something like that, and we have "allCount". Check EntityType code or EntityTypeDescriptor code for reference.

source/world/entity/MobSpawner.cpp
```diff
@@ -138,65 +136,51 @@ void MobSpawner::tick(Level *level, bool allowHostile, bool allowFriendly)
...
-
-    if (!level->isEmptyTile(pos)) 
-        return false;
-
-    if (!category->isFriendly() && brightness > MOB_SPAWNER_HOSTILE_BRIGHTNESS)
-        return false;
-
-    if (category->isFriendly()) 
-    {
-        if(brightness < MOB_SPAWNER_FRIENDLY_BRIGHTNESS || level->getTile(pos.below()) != TILE_GRASS)
```
BrentDaMage: Where did all this go?

source/world/level/Level.cpp
```diff
@@ -772,15 +772,22 @@ void Level::setTilesDirty(const TilePos& min, const TilePos& max)
 
 void Level::entityAdded(Entity* pEnt)
 {
+	EntityCategories::CategoriesMask mask = pEnt->getDescriptor().getCategories().getCategoryMask();
+	m_entityTypeCounts[mask]++;
```
BrentDaMage: You are indexing an array by a bitmask. You should be incrementing the count for each category that the entity is in. To do this, do a for loop to iterate over the `m_entityCountsByCategory` array, and temporarily bitshift the mask right by `i` on each iteration, then set `m_entityCountsByCategory[i]` to `bitshiftedMask & 1`.

source/world/level/Level.cpp
```diff
@@ -772,15 +772,22 @@ void Level::setTilesDirty(const TilePos& min, const TilePos& max)
...
 	{
 		LevelListener* pListener = *it;
 		pListener->entityAdded(pEnt);
 	}
+
 }
 
 void Level::entityRemoved(Entity* pEnt)
 {
+	EntityCategories::CategoriesMask mask = pEnt->getDescriptor().getCategories().getCategoryMask();
```
BrentDaMage, line 794: See above.

source/world/level/Level.cpp
```diff
@@ -1943,15 +1950,7 @@ float Level::getSunAngle(float f) const
 	return (float(M_PI) * getTimeOfDay(f)) * 2;
 }
 
-int Level::getEntityCountOfCategory(EntityCategories::CategoriesMask category) const
+int Level::countInstanceOfType(EntityCategories::CategoriesMask category)
```
BrentDaMage: Call this `getEntityCount`. That way we can just pass whatever we want to filter by as parameters.

source/world/level/Level.hpp
```diff
@@ -234,5 +235,7 @@ class Level : public LevelSource
 	int field_B10;
 	PathFinder* m_pPathFinder;
 	MobSpawner* m_pMobSpawner;
+
+	std::map<EntityCategories::CategoriesMask, int> m_entityTypeCounts;
```
BrentDaMage: Rename this to `m_entityCountsByCategory`

source/world/entity/MobCategory.hpp
```diff
@@ -6,6 +6,13 @@ class Material;
 class MobCategory
 {
 public:
+	enum ID {
```
BrentDaMage: Take inspiration from the EntityTypeDescriptors, since that's the most fleshed out version of we're basically already doing here.

BrentDaMage (changes requested):

source/world/entity/MobSpawner.cpp
rep-stosd: entity->canSpawn()

source/world/level/Level.cpp
rep-stosd: was working on resolving this but realise its more work that what its worth
1. your proposal doesn't work as intended as it doesn't count for all permutations of the hierarchy
2. the intended solution sucks
3. i misread the disassembly and thought so, but its actually EntityType::ID which has no hierarchy compared to this one

therefore we should keep it as is and hence `countInstanceOfType` as well

rep-stosd: in our case CategoriesMask serves as a raw tally of whatever belonging to parent EntityCategories, analogous to ::ID of other classes. we would pass EntityCategories itself to fetch the count rather than the mask, but it saves us from having to deal with polynomial time in one way or another

rep-stosd:
brother is there anything else to add?

source/world/entity/EntityCategories.cpp
```diff
@@ -1,5 +1,28 @@
...
+	ZOMBIE_MONSTER,  
+	ARTHROPOD,       
+	MINECART,        
+	SKELETON_MONSTER,
+	EQUINE_ANIMAL,   
+	PROJECTILE,      
+	ABSTRACT_ARROW,  
+	VILLAGER_BASE
+};
+const int EntityCategories::maskEnumCount = sizeof(EntityCategories::maskEnums) / sizeof(const EntityCategories::CategoriesMask);
```
BrentDaMage, line 24: `sizeof(const EntityCategories::CategoriesMask)` doesn't need the const.

source/world/entity/EntityCategories.cpp
```diff
@@ -1,5 +1,28 @@
 #include "EntityCategories.hpp"
 
+const EntityCategories::CategoriesMask EntityCategories::maskEnums[] =
```
BrentDaMage, line 3: Call this `all`

source/world/entity/EntityCategories.cpp
BrentDaMage, line 24: Call this `allCount`

BrentDaMage: Make this an unsigned int.

source/world/entity/MobCategory.cpp
```diff
@@ -5,12 +5,14 @@
...
 MobCategory MobCategory::creature      = MobCategory(EntityCategories(EntityCategories::ANIMAL),       10, 15, nullptr, true);
 MobCategory MobCategory::waterCreature = MobCategory(EntityCategories(EntityCategories::WATER_ANIMAL), 5,  10, nullptr, true);
-const MobCategory MobCategory::values[] = {
-	MobCategory::monster,
-	MobCategory::creature,
+
+const MobCategory* MobCategory::all[] = {
+	&MobCategory::monster,
+	&MobCategory::creature,
 	//MobCategory::waterCreature
```
BrentDaMage, line 12: Add ampersand within comment.

source/world/entity/MobFactory.cpp
```diff
@@ -59,9 +60,11 @@ void MobFactory::initMobLists() {
     waterCreatureList.insert(std::make_pair(EntityType::SQUID, 10));
 }
 
-const std::map<EntityType::ID, int>& MobFactory::GetMobListOfCategory(EntityCategories::CategoriesMask category) {
-    return category == EntityCategories::MONSTER ? monsterList :
-           category == EntityCategories::ANIMAL ? creatureList :
+const std::map<EntityType::ID, int>& MobFactory::GetMobListOfCategory(const EntityCategories& category) 
+{
+    EntityCategories::CategoriesMask mask = category.getCategoryMask();
+    return mask == EntityCategories::MONSTER ? monsterList :
```
BrentDaMage, line 66: Might be better as a map or a switch statement, just a suggestion.

BrentDaMage, line 66: You don't need to be fetching the mask manually, you can just compare the `EntitityCategories` to a mask directly.

source/world/level/Level.cpp
```diff
@@ -772,8 +772,14 @@ void Level::setTilesDirty(const TilePos& min, const TilePos& max)
 
 void Level::entityAdded(Entity* pEnt)
 {
+	// save for a bit
 	EntityCategories::CategoriesMask mask = pEnt->getDescriptor().getCategories().getCategoryMask();
```
BrentDaMage, line 776: Don't bother fetching the `CategoriesMask` directly, just compare the `EntityCategories` to a `CategoriesMask`.

source/world/level/Level.cpp
```diff
@@ -1950,7 +1961,7 @@ float Level::getSunAngle(float f) const
 	return (float(M_PI) * getTimeOfDay(f)) * 2;
 }
 
-int Level::countInstanceOfType(EntityCategories::CategoriesMask category)
+int Level::getEntityCount(const EntityCategories& category)
 {
-	return m_entityTypeCounts[category];
+	return m_entityCountsByCategory[category.getCategoryMask()];
```
BrentDaMage, line 1966: I'm still not sure how indexing by a bitmask would allow this to work correctly.

source/world/level/Level.hpp
```diff
@@ -195,7 +195,7 @@ class Level : public LevelSource
 	bool hasNeighborSignal(const TilePos& pos) const;
 
 
-	int countInstanceOfType(EntityCategories::CategoriesMask);
+	int getEntityCount(const EntityCategories&);
```
BrentDaMage, line 198: Make this return an unsigned int.

BrentDaMage (changes requested):

BrentDaMage:
Taken care of by #435, which forked this PR. Thank you guys.

PR #238, opened 2025-12-22
Don't redownload the iOS SDK if we don't have to

Un1q32:


Un1q32:
CI failed cuz github went down for a minute, everything works

PR #239, opened 2025-12-22
Fix compiler warns

Un1q32:
There's so many warn fixes at this point I can't keep track of them all.

Every warn that is not fixed by this PR (for the iOS build at least):
```
source/world/tile/TorchTile.cpp:162:10: warning: enumeration value 'DOWN' not handled in switch [-Wswitch]
  162 |         switch (face)
      |                 ^~~~
```
This looks like an actual bug, since the underside of torches are invisible ingame, I don't wanna silence the warn without fixing the bug, which is beyond the scope of this PR.

### DIFF
--- a/source/world/level/levelgen/chunk/RandomLevelSource.cpp
+++ b/source/world/level/levelgen/chunk/RandomLevelSource.cpp
@@ -378,69 +378,69 @@ void RandomLevelSource::postProcess(ChunkSource* src, const ChunkPos& pos)
 
 	for (int i = 0; i < 10; i++)
 	{
-		TilePos o(m_random.nextInt(16),
-		          m_random.nextInt(128),
-		          m_random.nextInt(16));
-		ClayFeature(Tile::clay->m_ID, 32).place(m_pLevel, &m_random, tp + o);
+		int xo = m_random.nextInt(16);
+		int yo = m_random.nextInt(128);
+		int zo = m_random.nextInt(16);
+		ClayFeature(Tile::clay->m_ID, 32).place(m_pLevel, &m_random, TilePos(tp.x + xo, yo, tp.z + zo));
 	}
 	
 	// Start of ore generation
 
 	for (int i = 0; i < 20; i++)
 	{
-		TilePos o(m_random.nextInt(16),
-		          m_random.nextInt(128),
-		          m_random.nextInt(16));
-		OreFeature(Tile::dirt->m_ID, 32).place(m_pLevel, &m_random, tp + o);
+		int xo = m_random.nextInt(16);
+		int yo = m_random.nextInt(128);
+		int zo = m_random.nextInt(16);
+		OreFeature(Tile::dirt->m_ID, 32).place(m_pLevel, &m_random, TilePos(tp.x + xo, yo, tp.z + zo));
 	}
 	for (int i = 0; i < 10; i++)
 	{
-		TilePos o(m_random.nextInt(16),
-		          m_random.nextInt(128),
-		          m_random.nextInt(16));
-		OreFeature(Tile::gravel->m_ID, 32).place(m_pLevel, &m_random, tp + o);
+		int xo = m_random.nextInt(16);
+		int yo = m_random.nextInt(128);
+		int zo = m_random.nextInt(16);
+		OreFeature(Tile::gravel->m_ID, 32).place(m_pLevel, &m_random, TilePos(tp.x + xo, yo, tp.z + zo));
 	}
 	for (int i = 0; i < 20; i++)
 	{
-		TilePos o(m_random.nextInt(16),
-		          m_random.nextInt(128),
-		          m_random.nextInt(16));
-		OreFeature(Tile::coalOre->m_ID, 16).place(m_pLevel, &m_random, tp + o);
+		int xo = m_random.nextInt(16);
+		int yo = m_random.nextInt(128);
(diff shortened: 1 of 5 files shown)


source/client/sound/SoundData.cpp
```diff
@@ -54,7 +54,7 @@ bool SoundDesc::_loadPcm(const AppPlatform* platform, const char *name)
     m_codecType = AudioCodec::PCM;
     m_fileData = m_file.data;
     m_header = *(PCMSoundHeader *) m_fileData;
-    m_buffer.m_pData = (void *) (m_fileData + sizeof(PCMSoundHeader));
+    m_buffer.m_pData = (short *) (m_fileData + sizeof(PCMSoundHeader));
```
BrentDaMage: Don't make this a short, make it an int16_t. We try to avoid shorts, longs, and long longs.

source/client/sound/SoundData.hpp
```diff
@@ -35,7 +35,7 @@ struct PCMSoundHeader
 
 struct SoundBuffer
 {
-	void* m_pData;
+	short *m_pData;
```
BrentDaMage: Make this int16_t*

source/client/sound/SoundStream.cpp
```diff
@@ -12,7 +12,7 @@ SoundStream::SoundStream()
     m_totalSamplesLeft = 0;
 
     m_tempPcmBuffer.m_dataSize = 4096 * 8;
-    m_tempPcmBuffer.m_pData = new int16_t[m_tempPcmBuffer.m_dataSize];
+    m_tempPcmBuffer.m_pData = new short[m_tempPcmBuffer.m_dataSize];
```
BrentDaMage: Make this int16_t

source/client/sound/SoundData.cpp
Un1q32: It's a short because the function its passed to from stb_vorbis or whatever takes a short*, I think it's better to keep the same type as where it ends up than to change it.

BrentDaMage: Nah, just cast it to an int16_t, that makes the size clearer on our end.

source/client/sound/SoundData.cpp
```diff
@@ -76,9 +76,7 @@ bool SoundDesc::_loadOgg(const AppPlatform* platform, const char* category, cons
     m_codecType = AudioCodec::OGG;
     m_fileData = m_file.data;
     m_header.m_bytes_per_sample = 2; // Always 2 (16-bit)
-    // Casting to a short** here might cause problems. Let's find out...
-    // Seems like it doesn't. Cool.
-    m_header.m_length = stb_vorbis_decode_memory(m_file.data, (int) m_file.size, &m_header.m_channels, &m_header.m_sample_rate, (short **) &m_buffer.m_pData);
+    m_header.m_length = stb_vorbis_decode_memory(m_file.data, (int) m_file.size, &m_header.m_channels, &m_header.m_sample_rate, (int16_t **) &m_buffer.m_pData);
```
BrentDaMage: We shouldn't have to cast this. Don't cast this, that way if someone changes the type of `SoundBuffer::m_pData`, the compiler will complain.

thirdparty/raknet/DynDNS.cpp
```diff
@@ -219,7 +219,7 @@ void DynDNS::Update(void)
 				existingHost[0]=0;
 				// Resolve DNS we are setting. If equal to current then abort
 				RakNetSocket2::DomainNameToIP(host.C_String(), existingHost);
-				if (existingHost && strcmp(existingHost, myIPStr)==0)
```
BrentDaMage, line 222: Why?

thirdparty/raknet/Rand.cpp
```diff
@@ -142,8 +142,8 @@ void seedMT( unsigned int seed, unsigned int *state, unsigned int *&next, int &l
 	// so-- that's why the only change I made is to restrict to odd seeds.
 	//
 
-	register unsigned int x = ( seed | 1U ) & 0xFFFFFFFFU, *s = state;
```
BrentDaMage, line 145: Just silence this warning if possible instead of removing the keyword.

BrentDaMage (changes requested):

thirdparty/raknet/DynDNS.cpp
Un1q32: It's stack allocated, so it can never be null.  The compiler warn is that it's always true.

thirdparty/raknet/Rand.cpp
rep-stosd: does register do anything meaningful in modern context?

Un1q32: I would have to check the warn again, but IIRC it basically that in C++98 it was deprecated and did nothing, and in some newer C++ version it was removed and won't compile, figured it was better to drop the keyword to support compiling for newer C++ standards.

Historically the keyword was used when compilers were worse at determining the hottest variables that should be put in a register, I think it can still be used like that in modern C, I've seen newer codebases use it, might just be inertia.  When I get the chance I'll check in compiler explorer if it makes a difference at all with any compiler.

BrentDaMage: Then go ahead and keep it removed.

BrentDaMage (approved):
LGTM

PR #241, opened 2025-12-28
Make ore and plant generation a bit more accurate

ljcool2006:
I found these fixes in [Wilyicaro's PEtoLE](https://github.com/Wilyicaro/PEtoLE) and have applied them to the current master.


BrentDaMage (approved):
LGTM

PR #242, opened 2025-12-31
Support building for C++98 on GLIBC using a vsnprintf wrapper.

Un1q32:
(v)snprintf requires C99, GLIBC refuses to declare it for lower standards. I don't think there's many devices that would be capable of running ReMCPE that don't have a C99 compliant libc, and we already kinda depend on C99 through things like C99 only printf format arguments.  My solution is just to declare a wrapper function in a C file that gets compiled with `-std=c99`.

MinGW must still be built for C++11 since their C++ headers don't support anything lower and intentionally error.

These projects should really be setup to detect all source files in directories automatically so it's not such a pain to add a new file.


BrentDaMage (changes requested):
All of the vsnprintf stuff should be moved to `compat/LegacyCPP_Compat.hpp`

compat/vsnprintf.c
```diff
@@ -0,0 +1,6 @@
+#include <stdio.h>
+
+int remcpe_vsnprintf(char *buf, size_t size, const char *format, va_list ap)
```
BrentDaMage, line 3: Put this in compat. Why not define it in the header?

BrentDaMage (changes requested):

compat/vsnprintf.c
Un1q32: It has to be in a C file so it can be compiled with `-std=c99` instead of `-std=c++98`, if we compile with `-std=c++98` GLIBC won't define vsnprintf.

BrentDaMage: In that case, have a vsnprintf.h and vspintf.c file in compat.

PR #246, opened 2026-01-04
Crash fix on iOS when shaders are enabled

Un1q32:
Makes the FEATURE_GFX_SHADERS macro public instead of private, also switches to using target_compile_definitions which is the way you're supposed to add macros in CMake.


BrentDaMage (changes requested):
Don't forget to do this for `D3D11` and `Null` too.

PR #248, opened 2026-01-04
More compiler warns

Un1q32:
Time to fix the -Wall warns

### DIFF
--- a/source/client/gui/screens/StartMenuScreen.cpp
+++ b/source/client/gui/screens/StartMenuScreen.cpp
@@ -542,6 +542,8 @@ void StartMenuScreen::draw2dTitle()
 {
 	Textures* tx = m_pMinecraft->m_pTextures;
 
+	// bool crampedMode = false;
+
 	currentShaderColor = Color::WHITE;
   
 	int yPos, width, height, left;
@@ -557,6 +559,7 @@ void StartMenuScreen::draw2dTitle()
 
 		if (m_width * 3 / 4 < m_2dTitleBounds.w)
 		{
+			// crampedMode = true;
 			yPos = 4;
 		}
 
@@ -585,6 +588,7 @@ void StartMenuScreen::draw2dTitle()
 	
 	      if (m_width * 3 / 4 < m_2dTitleBounds.w)
 	      {
+	        // crampedMode = true;
 	        yPos = 4;
 	      }
 	


compat/Unused.hpp
```diff
@@ -0,0 +1,7 @@
+#pragma once
+
+#ifdef __GNUC__
+#define REMCPE_UNUSED __attribute__((unused))
```
BrentDaMage: Please rename this to MC_UNUSED. We already have a couple macros for move function declarations that use the MC prefix, as well as VS and Xcode config var names. Plus, if we want to change the project name in the future, we won't have a bunch of old references.

platforms/android/AppPlatform_android.cpp
```diff
@@ -248,7 +248,6 @@ void AppPlatform_android::changeKeyboardVisibility(bool bShown)
 
 		// lInputMethodManager.hideSoftInput(...).
 		jmethodID MethodHideSoftInput = pEnv->GetMethodID(ClassInputMethodManager, "hideSoftInputFromWindow","(Landroid/os/IBinder;I)Z");
-		jboolean lResult = pEnv->CallBooleanMethod(lInputMethodManager, MethodHideSoftInput,lBinder, flags);
```
BrentDaMage, line 251: Why remove this? This looks like it does something.

platforms/android/main.cpp
```diff
@@ -43,31 +43,6 @@ struct engine
     NinecraftApp* ninecraftApp = nullptr;
 };
 
-static float mapStick(AInputEvent* event, int32_t axis)
```
BrentDaMage, line 46: Just comment these out, we might need them in the future.

platforms/sdl/sdl1/main.cpp
```diff
@@ -39,9 +39,9 @@ static void initGraphics()
     if (!mce::Platform::OGL::InitBindings())
     {
         const char* const GL_ERROR_MSG = "Error initializing GL extensions. OpenGL 2.0 or later is required.";
+        fputs(GL_ERROR_MSG, stderr);
```
BrentDaMage: Use LOG_E

source/client/gui/components/OptionList.cpp
```diff
@@ -340,6 +340,7 @@ void OptionList::initMultiplayerMenu()
 #ifdef __EMSCRIPTEN
 	m_items[currentIndex]->setDisabled(true);
 #endif
+	(void)currentIndex;
```
BrentDaMage: I understand why you did this, but please put comment explaining this.

source/client/gui/screens/StartMenuScreen.cpp
```diff
@@ -542,8 +542,6 @@ void StartMenuScreen::draw2dTitle()
 {
 	Textures* tx = m_pMinecraft->m_pTextures;
 
-	bool crampedMode = false;
```
BrentDaMage, line 545: Comment these out, since we have similar logic higher up that uses a hardcoded m_2dTitleBounds.w value.

source/client/player/input/MouseDevice.cpp
```diff
@@ -51,7 +52,7 @@ short MouseDevice::getY()
 
 bool MouseDevice::next()
 {
-	if (_index + 1 >= _inputs.size())
+	if ((size_t)_index + 1 >= _inputs.size())
```
BrentDaMage, line 55: Why not just change _index's time to size_t?

source/common/Mth.cpp
```diff
@@ -42,17 +43,19 @@ float Mth::invSqrt(float number)
...
 	const float threehalfs = 1.5F;
-
-	x2 = number * 0.5F;
-	y  = number;
-	i  = * ( int32_t * ) &y;                   // evil floating point bit level hacking
-	i  = 0x5f3759df - ( i >> 1 );              // what the fuck?
-	y  = * ( float * ) &i;
-	y  = y * ( threehalfs - ( x2 * y * y ) );  // 1st iteration
-  // y  = y * ( threehalfs - ( x2 * y * y ) );  // 2nd iteration, this can be removed
+	union {
```
BrentDaMage, line 48: Can't believe you modified the holy Quake math.

source/world/level/levelgen/chunk/RandomLevelSource.cpp
```diff
@@ -52,6 +52,7 @@ RandomLevelSource::RandomLevelSource(Level* level, int32_t seed, int version) :
 
 	Random random = m_random;
 	LOG_I("random.get : %d", random.genrand_int32() >> 1);
+	(void)random;
```
BrentDaMage: Also put comment here explaining why.

source/client/player/input/MouseDevice.cpp
Un1q32: It's initialized to -1

PR #250, opened 2026-01-05
Fixed building on MSVC

BrentDaMage:


Un1q32:
The actual fix for this is in the vsnprintf pr, add the sys/types.h include to that file

BrentDaMage:
> The actual fix for this is in the vsnprintf pr, add the sys/types.h include to that file

That doesn't fix it. As far as I can tell, ssize_t doesn't exist in MSVC.

BrentDaMage:
> > The actual fix for this is in the vsnprintf pr, add the sys/types.h include to that file
> 
> That doesn't fix it. As far as I can tell, ssize_t doesn't exist in MSVC.

Which is why we haven't used it anywhere in the project.

Un1q32:
uh just do a long then, and dont change the int32_t

PR #253, opened 2026-01-05
More compiler warns

Un1q32:
`-Wextra` enables some warns we don't want, so this just enables some parts we do want.
Also fixes 1 sign compare warn that was missed, and fixes a bug where a variable wasn't properly initialized.

### DIFF
--- a/thirdparty/zlib/gzguts.h
+++ b/thirdparty/zlib/gzguts.h
@@ -44,7 +44,7 @@
 #  define WIDECHAR
 #endif
 
-#ifdef WINAPI_FAMILY
+#if defined(WINAPI_FAMILY) || defined(_MSC_VER)
 #  define open _open
 #  define read _read
 #  define write _write


source/common/Logger.hpp
```diff
@@ -39,22 +40,60 @@ class Logger
...
-// TODO: Add a LoggerAndroid
-#define LOG(level, ...) __android_log_print(level, "ReMinecraftPE", __VA_ARGS__)
+	__android_log_vprint(loglevel, "ReMinecraftPE", fmt, ap);
 #else
-#define LOG(level, ...) Logger::singleton()->printf(level, __VA_ARGS__)
+	Logger::singleton()->vprintf(loglevel, fmt, ap);
 #endif
+}
+
+static inline void remcpe_log(enum eLogLevel loglevel, const char *fmt, ...)
```
BrentDaMage: Change the prefix to `mc_`

Un1q32:
note that after this pr log macro arguments are evaluated even in release builds, so be careful logging the results of expensive functions.

BrentDaMage:
> note that after this pr log macro arguments are evaluated even in release builds, so be careful logging the results of expensive functions.

Can you make sure this won't have any adverse effects to the current codebase?

Un1q32:
I ended up finding a way to not evaluate arguments on release builds, instead of `LOG_*("%s", function())` expanding to `mc_log("%s", function())` it expands to `if (0) mc_log("%s", function())`

BrentDaMage:
> I ended up finding a way to not evaluate arguments on release builds, instead of `LOG_*("%s", function())` expanding to `mc_log("%s", function())` it expands to `if (0) mc_log("%s", function())`

Great

source/common/Logger.hpp
```diff
@@ -39,22 +40,60 @@ class Logger
 
 #ifndef NDEBUG
 
+static inline void mc_vlog(enum eLogLevel loglevel, const char *fmt, va_list ap)
+{
 #ifdef ANDROID
-// TODO: Add a LoggerAndroid
```
BrentDaMage, line 43: Re-add TODO

BrentDaMage (changes requested):
One minor thing

BrentDaMage (approved):
LGTM

PR #259, opened 2026-01-10
add m_fAlpha logic for Buttons

Vimd0:
basic example:

m_testButton.m_fAlpha = 1.0f;


iProgramMC:
Should be m_fOpacity if it's a float, and not a boolean.

Wilyicaro:
I would say `m_fAlpha` is a better name, and is present on mojmap

Vimd0:
> Should be m_fOpacity if it's a float, and not a boolean.

oops

Vimd0:
> I would say `m_fAlpha` is a better name, and is present on mojmap

agreed

PR #260, opened 2026-01-10
Add credits screen

Un1q32:
Adds a credits screen accessible by clicking the logo on the title screen.
For now just a simple screen with some text and a few prominent contributors names in alphabetical order, no scrolling.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,7 +9,7 @@ set(CMAKE_C_STANDARD 99)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
 # Warnings
-add_compile_options(-Wall -Wignored-qualifiers -Wvariadic-macros)
+add_compile_options(-Wall -Wignored-qualifiers -Wvariadic-macros -pedantic -Wno-long-long)
 add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-Wc++11-extensions>)
 if(WERROR)
     add_compile_options(-Werror)
--- a/platforms/ios/minecraftpeAppDelegate.mm
+++ b/platforms/ios/minecraftpeAppDelegate.mm
@@ -36,7 +36,7 @@ NSError *G_audioSessionError = nil;
     {
         [self.window addSubview:self.viewController.view];
     }
-    NSLog(@"ViewController: %p\n", self.viewController);
+    NSLog(@"ViewController: %p\n", (void *)self.viewController);
     [self.window makeKeyAndVisible];
     return YES;
 }
@@ -63,7 +63,7 @@ NSError *G_audioSessionError = nil;
     {
         [application endBackgroundTask:bgTask];
         bgTask = UIBackgroundTaskInvalid;
-        NSLog(@"TIME RAN OUT %p\n", pthread_self()); // This originally used std::out, not sure why
+        NSLog(@"TIME RAN OUT %p\n", (void *)pthread_self()); // This originally used std::out, not sure why
     };
     bgTask = [application beginBackgroundTaskWithExpirationHandler:handler];
     viewController->suspended = YES;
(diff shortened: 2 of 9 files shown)


source/client/gui/screens/CreditsScreen.cpp
```diff
@@ -0,0 +1,99 @@
...
+CreditsScreen::~CreditsScreen()
+{
+    SAFE_DELETE(m_pDarkBackground);
+}
+
+void CreditsScreen::init()
+{
+	SAFE_DELETE(m_pDarkBackground);
+
+	m_pDarkBackground = new WorldSelectionList(m_pMinecraft, m_width, m_height);
```
TheBrokenRail: Why are you using an entire `WorldSelectionList` for the sole purpose of... drawing a dark background? That seems really over-complicated.

source/client/gui/screens/CreditsScreen.cpp
```diff
@@ -0,0 +1,99 @@
...
+void CreditsScreen::render(int mouseX, int mouseY, float f)
+{
+	renderBackground();
+	m_pDarkBackground->render(0, 0, f);
+
+	Screen::render(mouseX, mouseY, f);
+
+	drawCenteredString(*m_pMinecraft->m_pFont, "Credits", m_width / 2, 8, Color::WHITE);
+
+	std::istringstream credits(m_pMinecraft->m_pPlatform->readAssetFileStr("credits.txt", false));
```
TheBrokenRail: Reading the file from the FS every single frame seems excessive.

source/client/gui/screens/CreditsScreen.cpp
```diff
@@ -0,0 +1,99 @@
...
+	if (!credits) {
+		drawCenteredString(*m_pMinecraft->m_pFont, "Failed to load credits.txt", m_width / 2, 67, Color::WHITE);
+		return;
+	}
+
+	int height = 28;
```
TheBrokenRail: This should be properly centered IMO rather than using a hard-coded Y-offset.

source/client/gui/screens/CreditsScreen.cpp
BrentDaMage: Holy shit, do NOT do this.

Un1q32: oh is that what thats doing haha

source/client/gui/screens/CreditsScreen.cpp
BrentDaMage: If I recall correctly, there are two functions inside either the WorldSelectionList or both of the ScrolledSelectionLists that are responsible for drawing the top and bottom bars.

Un1q32: idk any other way to do it :P

source/client/gui/screens/CreditsScreen.cpp
```diff
@@ -0,0 +1,105 @@
...
+{
+	Screen::keyPressed(code);
+}
+
+void CreditsScreen::tick()
+{
+	m_pDarkBackground->tick();
+}
+
+static std::vector<std::string> credits;
```
TheBrokenRail: Rather than maintaining (and having to manually free) global state, why not just make this a property of `CreditsScreen`? Then just initialize it in the (obviously named) `CreditsScreen::init` function.

source/client/gui/screens/CreditsScreen.cpp
Un1q32: I have done it!

source/client/gui/screens/CreditsScreen.cpp
```diff
@@ -0,0 +1,105 @@
+/********************************************************************
```
BrentDaMage: Remove license

source/client/gui/screens/CreditsScreen.hpp
```diff
@@ -0,0 +1,32 @@
+/********************************************************************
```
BrentDaMage: Remove license

Un1q32:
Screenshot 2026-01-14 at 1 29 44 AM
image

BrentDaMage:
I was more wondering about the credits button itself on the menu.

Un1q32:
> I was more wondering about the credits button itself on the menu.

The main menu screen looks identical the before this PR because the button is invisible, but you can click on the logo to go to credits.

Un1q32:
A build with the opacity of the button set to 50% to visualize the click area
PXL_20260114_064624522

Un1q32:
Same thing on mac, the button texture breaks if it's stretched too long but it still functions if you click in the middle
Screenshot 2026-01-14 at 1 48 16 AM

BrentDaMage:
Cool

PR #276, opened 2026-01-14
Resource packs

### LINKED ISSUE #274: Resource Packs
We should be able to install resource packs in games/com.mojang/resource_packs to change textures/sounds/etc

Maybe implement bedrock's resource pack format so bedrock packs can be used as-is.  There's also beta 1.7.3's format but IIRC that one sucks so nah, maybe someone could make something to convert b1.7.3 packs automatically tho.

Un1q32:
First prototype of resource packs.
Resource packs are an overlay over the assets directory, stored in `games/com.mojang/resource_packs/packname`.
There is no UI for managing resource packs, you must manually edit options.txt.
You can enable a resource pack by setting `gfx_resourcepacks` to `packname`, you can enable multiple resource packs by setting it to `packname1,packname2,packname3`.

Other stuff:
Made m_externalStorageDir a property of AppPlatform.

Closes #274

### DIFF
--- a/platforms/audio/openal/SoundStreamOAL.hpp
+++ b/platforms/audio/openal/SoundStreamOAL.hpp
@@ -37,7 +37,7 @@ protected:
     bool _open(const std::string& fileName) override;
     void _close() override;
     void _update() override;
-    void _publishBuffer(unsigned int destBufferId, const SoundBuffer& sourceBuffer);
+    void _publishBuffer(unsigned int destBufferId, const SoundBuffer& sourceBuffer) override;
 
 private:
     std::vector<ALuint> m_buffers;
--- a/source/nbt/ListTag.hpp
+++ b/source/nbt/ListTag.hpp
@@ -61,7 +61,7 @@ public:
 	const CompoundTag* getCompound(unsigned int index) const;
 	Tag* copy() const override;
 	ListTag* copyList() const;
-	void deleteChildren();
+	void deleteChildren() override;
 
 	const std::vector<Tag*>& rawView() const { return m_list; }
 
--- a/source/world/entity/Arrow.hpp
+++ b/source/world/entity/Arrow.hpp
@@ -24,7 +24,7 @@ public:
 	void shoot(Vec3 pos, float speed, float r);
 	
 	void lerpMotion(float x, float y, float z) { lerpMotion(Vec3(x, y, z)); };
-	void lerpMotion(const Vec3& vel);
+	void lerpMotion(const Vec3& vel) override;
 
 	void tick() override;
 	void playerTouch(Player* pPlayer) override;
--- a/source/world/tile/DeadBush.hpp
+++ b/source/world/tile/DeadBush.hpp
@@ -7,5 +7,5 @@ class DeadBush : public Bush
 public:
 	DeadBush(int id, int texture);
 	int getResource(TileData, Random*) const override;
-	bool mayPlace(const Level*, const TilePos& pos) const;
+	bool mayPlace(const Level*, const TilePos& pos) const override;
 };


platforms/android/main.cpp
```diff
@@ -341,8 +341,8 @@ static void initWindow(struct engine* engine, struct android_app* app)
 
     if (!engine->initted)
     {
-        engine->ninecraftApp->m_externalStorageDir = getExternalStorageDir(engine);
-        g_AppPlatform.setExternalStoragePath(engine->ninecraftApp->m_externalStorageDir);
+        g_AppPlatform.m_externalStorageDir = getExternalStorageDir(engine);
```
BrentDaMage, line 344: Why twice?

platforms/ios/AppPlatform_iOS.h
```diff
@@ -32,11 +32,11 @@ class AppPlatform_iOS : public AppPlatform
 	int checkLicense() override;
 	int getScreenWidth() const override;
 	int getScreenHeight() const override;
-	void loadImage(ImageData& data, const std::string& path) override;
+	void loadImage(ImageData& data, const std::string& path, const std::vector<std::string>& = std::vector<std::string>()) override;
```
BrentDaMage: Do not have every single implementation of loadImage take a vector argument. Just have one version of loadImage in AppPlatform that takes the vector argument, and have it call the normal version of loadImage that is implemented in each AppPlatform.

platforms/ios/AppPlatform_iOS.mm
```diff
@@ -170,8 +172,20 @@
                 inDirectory: [NSString stringWithUTF8String:fileDir.c_str()]];
 }
 
-std::string AppPlatform_iOS::getAssetPath(const std::string &path) const
+std::string AppPlatform_iOS::getAssetPath(const std::string &path, const std::vector<std::string>& resourcepacks) const
 {
+	if (!resourcepacks.empty())
```
BrentDaMage: No, put this logic in some sort of base AppPlatform function.

platforms/ios/minecraftpeViewController.mm
```diff
@@ -190,7 +190,7 @@ - (void)viewDidLoad
     
     NSString *dir = [NSSearchPathForDirectoriesInDomains(NSDocumentDirectory, 1u, YES) objectAtIndex:0];
     NinecraftApp *app = new NinecraftApp();
-    app->m_externalStorageDir = [dir UTF8String];
+    platform->m_externalStorageDir = [dir UTF8String];
```
BrentDaMage, line 193: Use setExternalStoragePath here.

platforms/sdl/sdl1/main.cpp
```diff
@@ -283,9 +283,9 @@ int main(int argc, char* argv[])
 
     //LOG_I("Initializing AppPlatform...");
     g_pAppPlatform = new UsedAppPlatform(storagePath, screen);
+    g_pAppPlatform->m_externalStorageDir = storagePath;
```
BrentDaMage, line 286: Use setExternalStoragePath here.

platforms/sdl/sdl2/main.cpp
```diff
@@ -429,8 +429,8 @@ int main(int argc, char *argv[])
 	
 	// Start MCPE
 	g_pAppPlatform = new UsedAppPlatform(storagePath, window);
+	g_pAppPlatform->m_externalStorageDir = storagePath;
```
BrentDaMage, line 432: Use setExternalStoragePath here.

platforms/windows/main.cpp
```diff
@@ -154,7 +154,7 @@ int WINAPI WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPSTR lpCmdLine
 
 	g_pApp = new NinecraftApp;
 	g_pApp->m_pPlatform = &g_AppPlatform;
-	g_pApp->m_externalStorageDir = ".";
+	g_AppPlatform.m_externalStorageDir = ".";
```
BrentDaMage: Use setExternalStoragePath here.

source/client/app/AppPlatform.cpp
```diff
@@ -323,28 +329,33 @@ SoundSystem* AppPlatform::getSoundSystem() const
...
-	if (realPath.size() && realPath[0] == '/')
+	if (!resourcepacks.empty())
 	{
-		// trim it off
-		realPath = realPath.substr(1);
+		for (size_t i = 0; i < resourcepacks.size(); ++i)
+		{
+			std::string fullpath = getAssetPath("/resource_packs/" + resourcepacks[i] + "/" + path);
+			struct stat st;
+			if (stat(fullpath.c_str(), &st) == 0 && S_ISREG(st.st_mode))
```
BrentDaMage: There is no way we need to have the same code copy-pasted in 6 different platform-specific implementations. Abstract this into some function inside the AppPlatform base class. All of our file operations thus far have been via the standard library or functions in Utils.hpp.

source/client/app/NinecraftApp.cpp
```diff
@@ -172,14 +172,17 @@ void NinecraftApp::_initAll()
...
+	m_pLevelStorageSource = new ExternalFileLevelStorageSource(platform()->m_externalStorageDir);
 #endif
 
 	m_pGui = new Gui(this);
 	// "Default.png" for the launch image overwrites "default.png" for the font during app packaging
 	std::string font = "font/default8.png";
-	if (!platform()->doesTextureExist(font))
-		font = "font/default.png";
+	{
+		AssetFile file = platform()->readAssetFile(font, true, getOptions()->m_resourcepacks);
```
BrentDaMage: We really should not have to pass the resource packs every single time we call readAssetFile. This is stupid.

source/client/options/Options.cpp
```diff
@@ -292,6 +324,7 @@ std::vector<std::string> Options::getOptionStrings()
 	SO("misc_oldtitle",             saveBool(m_b2dTitleLogo));
 	SO("info_debugtext",            saveBool(m_bDebugText));
 	SO("misc_menupano",			    saveBool(m_bMenuPanorama));
+	SO("gfx_resourcepacks",		    saveArray(m_resourcepacks));
```
BrentDaMage: Does this format match Java's?

source/client/renderer/renderer/RenderMaterialGroup.cpp
```diff
@@ -69,18 +70,18 @@ struct MaterialTreePopulator
...
     }
 
-	void operator()(const std::string& materialName, MaterialParent& materialParent)
+	void operator()(const std::string& materialName, MaterialParent& materialParent, const std::vector<std::string>& resourcepacks)
 	{
         // Construct material with parent and place in group
         RenderMaterial& parentMaterial = group._getMaterialOrDefault(materialParent.parentName, groupBaseParent);
         const rapidjson::Value::ConstObject obj = materialParent.json->GetObj();
-        group.m_materials[materialName] = RenderMaterial(obj, parentMaterial);
+        group.m_materials[materialName] = RenderMaterial(obj, parentMaterial, resourcepacks);
```
BrentDaMage: Polluting the rendering code with this resourcepack shit. Do not do this.

source/client/sound/SoundData.cpp
```diff
@@ -30,22 +30,43 @@ bool SoundDesc::_load(const AppPlatform* platform, const char* category, const c
 	}
 
     // Load
-    if (_loadOgg(platform, category, name) || _loadPcm(platform, name))
+	std::vector<std::string> resourcepacks = options->m_resourcepacks;
```
BrentDaMage: You're copying the vector here.

source/client/sound/SoundData.cpp
```diff
@@ -30,22 +30,43 @@ bool SoundDesc::_load(const AppPlatform* platform, const char* category, const c
...
+	std::string packdir, slashname = "/" + std::string(name);
+	bool ret = false;
+
+	// try to use the resource pack version of the texture
+	for (size_t i = 0; i < resourcepacks.size(); ++i)
+	{
+        packdir = "/resource_packs/" + resourcepacks[i] + "/";
```
BrentDaMage: Dude, we have this logic in the 6 different AppPlatform implementations, and here? At the very least, the path should be some sort of macro, and not be copy-pasted every single time it's used.

source/client/sound/SoundData.cpp
```diff
@@ -111,10 +128,10 @@ void SoundDesc::_unload()
 }
 
 // Load All Sounds
-void SoundDesc::_loadAll(const AppPlatform* platform)
+void SoundDesc::_loadAll(const AppPlatform* platform, Options* options)
 {
-#define SOUND(category, name) SA_##name._load(platform, #category, #name);
-#define SOUND_NUM(category, name, number) SA_##name##number._load(platform, #category, #name#number);
+#define SOUND(category, name) SA_##name._load(options, platform, #category, #name);
```
BrentDaMage: Do not do this, find a better way.

BrentDaMage: Pass a pointer to Minecraft or something, that way it can call getOptions() and platform()

source/common/utility/InheritanceTree.hpp
```diff
@@ -48,7 +48,7 @@ class InheritanceTree
     }
 
     template <class TVisitor>
-    void visitBFS(TVisitor& visitor)
+    void visitBFS(TVisitor& visitor, const std::vector<std::string>& resourcepacks)
```
BrentDaMage: Dude no. Absolutely not. Do not pollute shared utility code with resourcepack stuff. Never do this. If you have questions during implementation, please ask me.

source/renderer/ShaderGroup.cpp
```diff
@@ -61,7 +61,7 @@ ShaderProgram& ShaderGroup::getShaderProgram(ShaderType shaderType, const std::s
     std::string programCode, shaderPath;
     if (Util::isValidPath(codeOrPath))
     {
-        programCode = AppPlatform::singleton()->readAssetFileStr(codeOrPath, true);
+        programCode = AppPlatform::singleton()->readAssetFileStr(codeOrPath, true, resourcepacks);
```
BrentDaMage: To get a better idea as to how this can be done cleanly, we should check what PE is doing.

BrentDaMage (changes requested):

source/client/app/NinecraftApp.cpp
Un1q32: The only other way would be a function that already knows what resource packs are active because it can access options, which would be impossible from AppPlatform, putting it in Minecraft or something else would mean rewriting a fuckton of code so it can have access to a Minecraft*, I tried this already trust me the simplest way to do this is just to pass the array.

source/client/options/Options.cpp
Un1q32: No idea

source/client/sound/SoundData.cpp
Un1q32: It needs to be duplicated in SoundData because of how the sound loading code works.
Because the sound loading code tries to load ogg files before pcm files, if we have 2 resource packs, pack1 and pack2, pack1 is at the top, and pack2 is below, and both replace the sound foo, but pack1 stores their version of foo as a pcm and pack2 stores it as an ogg, the game will use pack2's version if we don't duplicate the logic. The user expects pack1's version of the sound to be used because it is before pack2 in the list.

source/renderer/ShaderGroup.cpp
Un1q32: PE only added resource packs in 0.15 which is after they removed symbols, I'm not decomping that.

platforms/windows/main.cpp
Un1q32: setExternalStoragePath is a function specific to android and doesn't even do the same thing

platforms/sdl/sdl2/main.cpp
Un1q32: setExternalStoragePath is a function specific to android and doesn't even do the same thing

platforms/sdl/sdl1/main.cpp
Un1q32: setExternalStoragePath is a function specific to android and doesn't even do the same thing

platforms/ios/minecraftpeViewController.mm
Un1q32: setExternalStoragePath is a function specific to android and doesn't even do the same thing

platforms/android/main.cpp
Un1q32: idk man ask the guy who wrote the code originally

source/client/options/Options.cpp
BrentDaMage: I'm assuming, based on the name, that you can apply multiple resource packs at once. Is this correct?

Un1q32: yes, for more than one just use commas, `pack1,pack2` makes pack2 get used as a fallback for any textures that aren't in pack1 before the final fallback to vanilla textures.

source/client/sound/SoundData.cpp
Un1q32: It also can't be a macro because the functions that are called inside the for loop have to change.

Un1q32: This logic is now only duplicated once, that's as good as I can get it I think.

platforms/ios/AppPlatform_iOS.mm
```diff
@@ -172,6 +172,8 @@
 
 std::string AppPlatform_iOS::getAssetPath(const std::string &path) const
 {
+	if (path.size() && path[0] == '/')
```
BrentDaMage: Use `!path.empty()`

platforms/ios/AppPlatform_iOS.mm
```diff
@@ -172,6 +172,8 @@
...
+		return m_externalStorageDir + "/games/com.mojang" + path;
```
BrentDaMage: Rename all implementations of `AppPlatform::getAssetPath()` to `AppPlatform::_getAssetPath()`, and make the methods `protected` instead of `public`. Then, have a `AppPlatform::getAssetPath()` function in the base class that uses this logic, and calls `_getAssetPath()` otherwise. That way, we are never repeating the logic.

source/client/app/AppPlatform.cpp
```diff
@@ -148,10 +149,15 @@ void AppPlatform::loadImage(ImageData& data, const std::string& path)
 	data.m_colorSpace = channels == 3 ? COLOR_SPACE_RGB : COLOR_SPACE_RGBA;
 }
 
-TextureData AppPlatform::loadTexture(const std::string& path, bool bIsRequired)
+void AppPlatform::loadImageResource(ImageData& data, const std::string& path, const std::vector<std::string>& resourcepacks)
```
BrentDaMage: Actual PE calls `Resource::loadTexture` so that they never have to pollute the `AppPlatform` class with game feature logic.

source/client/app/AppPlatform.cpp
```diff
@@ -325,15 +331,25 @@ SoundSystem* AppPlatform::getSoundSystem() const
...
 std::string AppPlatform::getAssetPath(const std::string& path) const
 {
-	std::string realPath = path;
-	if (realPath.size() && realPath[0] == '/')
+	if (path.size() && path[0] == '/')
+		return m_externalStorageDir + "/games/com.mojang" + path;
+	return "assets/" + path;
+}
+
+std::string AppPlatform::getResourcePath(const std::string& path, const std::vector<std::string>& resourcepacks) const
```
BrentDaMage: Everything else in actual PE uses `Resource::load`, including `mce::RenderMaterialGroup` and `mce::ShaderGroup`.

BrentDaMage: Honestly, I think all of our asset/resource loading functions should be getting called in a separate class like `Resource`, `AppPlatform` should be solely responsible for getting file contents and paths.

source/client/app/NinecraftApp.cpp
```diff
@@ -62,13 +62,13 @@ void NinecraftApp::_initTextures()
 
 	if (GrassColor::isAvailable())
 	{
-        TextureData textureData = m_pPlatform->loadTexture("misc/grasscolor.png", true);
-		GrassColor::init(textureData);
+        TextureData* textureData = m_pTextures->loadTexture("misc/grasscolor.png", true);
```
BrentDaMage: Good change. This would not have survived us losing the render context otherwise.

source/client/options/Options.cpp
```diff
@@ -188,6 +190,21 @@ int Options::readInt(const std::string& str)
 	return f;
 }
 
+std::vector<std::string> Options::readPackArray(const std::string& str)
```
BrentDaMage: std::vector should be passed in as a reference parameter, instead of being a return value. That way we avoid having to copy altogether.

source/client/options/Options.cpp
```diff
@@ -188,6 +190,21 @@ int Options::readInt(const std::string& str)
...
+{
+	std::vector<std::string> ret;
+	std::istringstream ss(str);
+	std::string pack;
+
+	while (std::getline(ss, pack, ','))
+	{
+		std::string packpath = AppPlatform::singleton()->getAssetPath("/resource_packs/" + pack);
```
BrentDaMage: `/resource_packs/` path should be a global macro somewhere.

source/client/options/Options.cpp
```diff
@@ -200,6 +217,25 @@ std::string Options::saveInt(int i)
 	return ss.str();
 }
 
+std::string Options::saveArray(std::vector<std::string> arr)
```
BrentDaMage: You forgot to pass the parameter by reference here.

source/client/options/Options.cpp
BrentDaMage: Great. Seems solid.

source/client/options/Options.hpp
```diff
@@ -72,8 +72,10 @@ class Options
 private:
 	static bool readBool(const std::string& str);
 	static int readInt(const std::string& str);
+	static std::vector<std::string> readPackArray(const std::string& str);
```
BrentDaMage: Have a `readArray` function. Have `readPackArray` call this function, then iterate back over the results and apply its path transformations.

source/client/options/Options.hpp
```diff
@@ -131,6 +133,7 @@ class Options
 	bool m_bDynamicHand;
 	bool m_b2dTitleLogo;
 	bool m_bMenuPanorama;
+	std::vector<std::string> m_resourcepacks;
```
BrentDaMage: should be `m_resourcePacks`

source/client/sound/SoundData.cpp
```diff
@@ -30,22 +30,42 @@ bool SoundDesc::_load(const AppPlatform* platform, const char* category, const c
...
 
     // Load
-    if (_loadOgg(platform, category, name) || _loadPcm(platform, name))
+	std::string packdir, slashname = "/" + std::string(name);
+	bool ret = false;
+
+	// try to use the resource pack version of the sound
+	for (size_t i = 0; i < resourcepacks.size(); ++i)
+	{
+        packdir = "/resource_packs/" + resourcepacks[i] + "/";
```
BrentDaMage: I'm really not a fan of having resource pack logic here and in AppPlatform. It really should just exist in one place, ideally `Resource`. Make a singleton class called `Resources` if you have to, and have everything call that instead of `AppPlatform` directly.

source/common/Utils.cpp
```diff
@@ -182,6 +186,22 @@ bool DeleteDirectory(const std::string& name2, bool unused)
 #endif
 }
 
+bool isRegularFile(const char *path)
```
BrentDaMage, line 190: What is a "regular file"?

source/client/app/NinecraftApp.cpp
Un1q32: Had to anyway because AppPlatform's loadTexture doesn't support resourcepack logic, but this was def a bug before, AppPlatform's loadTexture doesn't support the isRequired argument either, the logic is handled in the Textures' loadTexture and ignored in AppPlatform, I removed the argument altogether to prevent future confusion, and also for the small performance improvement of having 1 less argument to pass.

source/common/Utils.cpp
Un1q32: whatever `S_ISREG` is successful on, for my purposes it just means not a directory or special file like those found in /dev.

Un1q32: Screenshot 2026-01-24 at 4 27 41 PM

From stackoverflow, note that in this case, if your path points to a symlink the result will be whatever the target of the symlink is, not the symlink itself, since I used the normal stat and not lstat or anything like that, which follows symlinks.

platforms/ios/AppPlatform_iOS.mm
```diff
@@ -146,7 +146,7 @@
 	return true;
 }
 
-NSString* AppPlatform_iOS::_getAssetPath(const std::string &path) const
+NSString* AppPlatform_iOS::__getAssetPath(const std::string &path) const
```
BrentDaMage: Rename this to `_getBundleResourcePath`

source/client/app/AppPlatform.hpp
```diff
@@ -127,7 +128,7 @@ class AppPlatform
 	virtual void initSoundSystem();
 	virtual SoundSystem* getSoundSystem() const;
 	// Used For Sounds
-	virtual std::string getAssetPath(const std::string& path) const;
+	std::string getAssetPath(const std::string& path) const;
```
BrentDaMage: For the sake of consistency, make this virtual as well.

PR #277, opened 2026-01-14
Fix menu performance on macOS arm64

Un1q32:
Enables the GLES1_WORKAROUND on arm64 macOS

### DIFF
--- a/source/client/gui/screens/StartMenuScreen.cpp
+++ b/source/client/gui/screens/StartMenuScreen.cpp
@@ -420,6 +420,7 @@ StartMenuScreen::StartMenuScreen() :
 {
 	m_pTiles = nullptr;
 	m_chosenSplash = -1;
+	m_p2dTitleTex = nullptr;
 	m_bUsingJavaLogo = false;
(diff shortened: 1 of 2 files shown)


source/renderer/hal/ogl/BufferOGL.cpp
```diff
@@ -127,7 +127,7 @@ void BufferOGL::updateBuffer(RenderContext& context, unsigned int stride, void*&
     // to be rendered, so perhaps some day...
     // Additionally, we could try holding the vertex buffer data in memory and pass
     // it in the draw call, as supposedly not even using buffers is faster.
-#ifdef GL_VERSION_ES_CM_1_0
```
BrentDaMage, line 130: Why did you remove this???
This issue only occurs on GLES 1 on iOS and Android, not GLES2. We need this check here!

Un1q32: It occurs on Android?  I thought it only happened with the Apple GPU driver.

PR #278, opened 2026-01-15
Fix credits when credits.txt is DOS format.

Un1q32:


source/client/gui/screens/CreditsScreen.cpp
```diff
@@ -28,7 +28,11 @@ void CreditsScreen::init()
 
 	std::string line;
 	while (std::getline(credits_stream, line))
+	{
+		if (line[line.size() - 1] == '\r')
```
TheBrokenRail: As I mentioned on Discord, this should also check if the line is empty first.

PR #286, opened 2026-01-19
remove unnessisary casting to int

Un1q32:
There's a lot of places where we cast the result of .size() in a for loop to int for no reason, maybe mojang did this to silence sign compare warns but they were stupid about how they did it. Using size_t directly is more correct and probably a little faster.


.clangd
```diff
@@ -0,0 +1,2 @@
+CompileFlags:
+  Add: [-I/home/uniq/devel/mcpe/source, -I/home/uniq/devel/mcpe, -I/home/uniq/devel/mcpe/thirdparty/raknet, -I/home/uniq/devel/mcpe/thirdparty/stb_image/include, -isystem/usr/include/SDL2]
```
BrentDaMage: mistake?

Un1q32: Yeah that was a mistake, added to .gitignore so it doesn't happen again.

PR #295, opened 2026-01-24
Nintendo Switch support

Bagietas:
Everything seems to work just fine.
Currently it's not compileable without commenting out `set(CMAKE_CXX_EXTENSIONS OFF)` in _CMakeLists.txt_

### DIFF
--- a/platforms/sdl/sdl2/CMakeLists.txt
+++ b/platforms/sdl/sdl2/CMakeLists.txt
@@ -90,6 +90,6 @@ if(NINTENDO_SWITCH)
 
     nx_create_nro(reminecraftpe
         NACP reminecraftpe.nacp
-        ICON "${MC_ROOT}/game/assets/icon.jpg"
+        ICON "../../../game/assets/icon.jpg"
     )
 endif()


Bagietas:
It is compileable now (reference to PR description)

.github/workflows/build.yml
```diff
@@ -175,3 +175,20 @@ jobs:
...
+          - name: Checkout Repository
+            uses: actions/checkout@v3
+            with:
+                submodules: true
+
+          - name: Build Switch binary
+            run: |
+                mkdir build
+                cd build
+                /opt/devkitpro/portlibs/switch/bin/aarch64-none-elf-cmake -GNinja -DWERROR=ON ..
```
BrentDaMage, line 193: Shouldn't we be using the same flags that we use for the standard CMake builds?

platforms/sdl/sdl2/CMakeLists.txt
```diff
@@ -77,3 +80,16 @@ if(EMSCRIPTEN) # WASM
...
+if(NINTENDO_SWITCH)
+    nx_generate_nacp(reminecraftpe.nacp
+        NAME "ReMinecraftPE"
+        AUTHOR "ReMinecraftPE Team"
+        VERSION "v0.2.0 alpha"
+    )
+
+    nx_create_nro(reminecraftpe
+        NACP reminecraftpe.nacp
+        ICON "../../../game/assets/icon.jpg"
```
BrentDaMage: I think we may have an `MC_ROOT` variable you can use for this. If not, this is fine.

.github/workflows/build.yml
Un1q32: (i made the CI bit) aren't these already the same flags, the only thing missing is the matrix thing, and that seems to be a way to reuse the same CI script but slightly modified with slightly tweaked flags, but we only have 1 switch build so we don't need it.

Bagietas:
Already existing OpenGL + SDL implementation was done just fine, so it only required fixing compiler and linker errors.

platforms/sdl/sdl2/CMakeLists.txt
Bagietas: Can't use that macro there.

BrentDaMage: Yeah I didn't even know if it actually existed.

Bagietas: Actually you might want to look into android and ios builds as they are using that macro to check if some asset is present, possibly breaking some.

BrentDaMage: That macro exists in Xcode and VS, just not in CMake.

Bagietas: Got it.

PR #296, opened 2026-01-25
Copyright-free assets

Un1q32:
Adds new textures that are redistributable

Requires #310

### DIFF
--- a/game/assets/readme.txt
+++ b/game/assets/readme.txt
@@ -3,9 +3,9 @@ GGO4: parts of items.png, parts of terrain.png, cow.png
 glitchy8015: default_world.png
 Goodly: parts of terrain.png, all mobs except the cow, All of gui.png except touch buttons, clouds.png, font/default.png, p_rose.png, parts of p_crying_obsidian.png
 iProgramMC: parts of terrain.png, parts of p_crying_obsidian.png
-Un1q32: parts of terrain.png, particles.png, icons.png, parts of items.png, p_camera.png
+Un1q32: parts of terrain.png, particles.png, parts of items.png, p_camera.png
 Wilylcaro: gui.png touch buttons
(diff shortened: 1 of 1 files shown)


game/assets/gui/gui.png
BrentDaMage, line 1: The normal buttons look like they're squashed horizontally. Is this intended?

Un1q32: looks fine to me?
Screenshot 2026-01-25 at 7 55 23 PM
Screenshot 2026-01-25 at 7 55 40 PM

BrentDaMage: Do you see how it looks like someone took the insides of the texture and squashed it? Or stretched it really tall vertically?

Un1q32: oh yea i think thats just the style

Un1q32: its straight from clasicube so

game/assets/terrain.png
BrentDaMage, line 1: Could you put a slight outline around the repeaters?

tools/extract_apk.py
```diff
@@ -39,7 +40,7 @@ def extract_sounds(apk_zip):
 # Extract Assets
 def extract_assets(apk_zip):
     assets = os.path.dirname(os.path.abspath(__file__))
-    assets = os.path.join(assets, '..', 'game', 'assets')
+    assets = os.path.join(assets, '..', 'resource_pack')
```
BrentDaMage: Could we possibly add support for using resource packs that are packaged with the app? For example, instead of having "resource_packs" in the external storage dir, we search inside of "assets/resource_packs". PE bakes a couple vanilla-related resource packs into the assets folder like this, and I feel like it'd fit perfectly here.

Un1q32: I'll need you to add support for actually using resource packs from that directory.

BrentDaMage (approved):

PR #314, opened 2026-01-30
Pickaxe, Hatchet, & Shovel

samdotjpg:
Up for discussion on how to implement the code for tool compatibility when breaking tiles. Thanks!

Check Item.cpp's Item Tier definitions to uncomment the durabilities from zero. This was done so tools don't just break since there's no crafting.


source/client/renderer/entity/ItemRenderer.cpp
```diff
@@ -189,9 +189,30 @@ void ItemRenderer::renderGuiItemOverlay(Font* font, Textures* textures, ItemStac
...
+		int duraColor = ((255 - duraPercent) << 16) | (duraPercent << 8);
+
+		Tesselator& t = Tesselator::instance;
+		
+		blitRect(t, x + 2, y + 13, 13, 2, 0);
+		blitRect(t, x + 2, y + 13, 12, 1, duraBgColor);
+		blitRect(t, x + 2, y + 13, duraWidth, 1, duraColor);
+
+
+		glColor4f(1.0F, 1.0F, 1.0F, 1.0F);
```
BrentDaMage: Remove this

source/world/item/Item.cpp
```diff
@@ -800,10 +802,10 @@ Item
 	*Item::quiver;
 
 Item::Tier
-	Item::Tier::WOOD   (0, 59,   2.0f,  0),
-	Item::Tier::STONE  (1, 131,  4.0f,  1),
-	Item::Tier::IRON   (2, 250,  6.0f,  2),
-	Item::Tier::EMERALD(3, 1561, 8.0f,  3),
-	Item::Tier::GOLD   (0, 32,   12.0f, 0);
+	Item::Tier::WOOD   (0, 0/* 59 */,   2.0f,  0),
```
BrentDaMage: Why are `uses` set to 0?

source/world/item/ItemStack.hpp
```diff
@@ -64,22 +64,22 @@ class ItemStack
...
 	void hurtAndBreak(int, Entity*);
-	void hurtEnemy(Mob*);
+	void hurtEnemy(Mob*, Player*);
 	void interactEnemy(Mob*);
 	bool isDamageableItem() const;
 	bool isDamaged() const;
 	bool isStackable() const;
 	bool isStackedByData() const;
-	void mineBlock(const TilePos& pos, Facing::Name face);
+	void mineBlock(const TilePos& pos, Facing::Name face, Player* player);
```
BrentDaMage: Why are we passing a Player ptr here? They don't seem to do this on PE or Java.

source/world/item/PickaxeItem.cpp
```diff
@@ -0,0 +1,59 @@
+#include "PickaxeItem.hpp"
+#include "world/tile/Tile.hpp"
+
+PickaxeItem::PickaxeItem(int id, Tier& tier) : ToolItem(id, 1, tier)
+{
+   static const Tile* pickaxeTiles[] =
+   {
+      Tile::stoneBrick,
```
BrentDaMage: I'm really not a fan of this. I think we should look at how this is being done for newer versions, or in PE. Perhaps we have some sort of hardness value (required tool tier) and tool-type bitmask inside each Tile definition?

source/world/level/Material.cpp
```diff
@@ -133,6 +138,12 @@ bool Material::isFlammable() const
 	return m_bFlammable;
 }
 
+Material* Material::setNonMineable()
```
BrentDaMage: In 0.12.1, this is called `setNotAlwaysDestroyable`

BrentDaMage (changes requested):

source/world/item/Item.cpp
samdotjpg: Read above comment on the original PR please

source/client/renderer/entity/ItemRenderer.cpp
samdotjpg: All of the rendering or just the color reset?

source/world/item/ItemStack.hpp
samdotjpg: Should I pass in a nullptr to hurtAndBreak instead? It needs an entity for the future item break stat. Please let me know what you would prefer instead to get this data for future reference.

source/client/renderer/entity/ItemRenderer.cpp
BrentDaMage: Just the direct GL calls. That's likely the reason why the Direct3D11 build is failing.

samdotjpg: Gotcha, resolved

source/world/item/Item.hpp
```diff
@@ -72,16 +73,16 @@ class Item
...
-	virtual float getDestroySpeed(ItemStack*, Tile*) const;
+	virtual float getDestroySpeed(ItemStack*, const Tile*) const;
 	virtual ItemStack* use(ItemStack*, Level*, Player*) const;
 	virtual int getMaxStackSize() const;
 	virtual TileData getLevelDataForAuxValue(int x) const;
 	virtual bool isStackedByData() const;
 	virtual int getMaxDamage() const;
 	virtual void hurtEnemy(ItemStack*, Mob*) const;
-	virtual void mineBlock(ItemStack*, const TilePos& pos, Facing::Name face) const;
+	virtual void mineBlock(ItemStack*, const TilePos& pos, Facing::Name face, Player* player) const;
```
BrentDaMage: This should take a Mob*, not a Player*

source/world/item/ItemStack.hpp
BrentDaMage: Looks like in PE, this takes a Mob*

source/world/item/ItemStack.hpp
```diff
@@ -64,22 +64,22 @@ class ItemStack
...
 	std::string getHovertextName() const;
-	float getDestroySpeed(Tile*);
+	float getDestroySpeed(const Tile*);
 	int getIcon() const;
 	int getMaxDamage() const;
 	int getMaxStackSize() const;
 	void hurt(int by);
 	void hurtAndBreak(int, Entity*);
-	void hurtEnemy(Mob*);
+	void hurtEnemy(Mob*, Player*);
```
BrentDaMage: This should just take two Mob ptrs

BrentDaMage (changes requested):
Most of the Player ptr parameters you added should probably just become Mob ptrs.

source/world/item/Item.hpp
```diff
@@ -45,17 +45,18 @@ class Item
...
-			m_Durability(b),
-			m_HitStrength(c),
-			field_C(d)
-		{}
+		int   m_level;
+		int   m_uses;
+		float m_speed;
+		int   m_damage;
+
+		Tier(int level, int uses, float speed, int damage) :
```
BrentDaMage, line 53: Here's code from PE 0.12.1:

```cpp
ItemInstance *__fastcall Item::Tier::getTierItem(ItemInstance *retval, Item::Tier *this)
{
  ItemInstance *result; // r0

  switch ( this->m_level )
  {
    case 0:
      if ( this->m_speed == 2.0 )
        ItemInstance::ItemInstance(retval, (const Tile *)Tile::wood);
      else
        ItemInstance::ItemInstance(retval, (const Item *)Item::goldIngot);
      result = retval;
      break;
    case 1:
      ItemInstance::ItemInstance(retval, (const Tile *)Tile::rock);
      result = retval;
      break;
    case 2:
      ItemInstance::ItemInstance(retval, (const Item *)Item::ironIngot);
      result = retval;
      break;
    case 3:
      ItemInstance::ItemInstance(retval, (const Item *)Item::diamond);
      result = retval;
      break;
    default:
      ItemInstance::ItemInstance(retval);
      result = retval;
      break;
  }
  return result;
}
```

It's used for `Item::isValidRepairItem()`. I guess we won't really need it then.

source/world/item/Item.cpp
```diff
@@ -802,10 +801,10 @@ Item
...
-	Item::Tier::WOOD   (0, 0/* 59 */,   2.0f,  0),
-	Item::Tier::STONE  (1, 0/* 131 */,  4.0f,  1),
-	Item::Tier::IRON   (2, 0/* 250 */,  6.0f,  2),
-	Item::Tier::EMERALD(3, 0/* 1561 */, 8.0f,  3),
-	Item::Tier::GOLD   (0, 0/* 32 */,   12.0f, 0);
+	Item::Tier::WOOD   	(0, 59,   2.0f,  0),
+	Item::Tier::STONE  	(1, 131,  4.0f,  1),
+	Item::Tier::IRON   	(2, 250,  6.0f,  2),
+	Item::Tier::EMERALD	(3, 1561, 8.0f,  3),
+	Item::Tier::GOLD   	(0, 32,   12.0f, 0);
```
samdotjpg, line 808: I ultimately left the durabilities in check for this PR but as you know they're easy enough to take away.

BrentDaMage: Keep them, it's fine.

source/client/renderer/entity/ItemRenderer.cpp
```diff
@@ -191,9 +191,27 @@ void ItemRenderer::renderGuiItemOverlay(Font* font, Textures* textures, ItemStac
 	if (item.isEmpty())
 		return;
 
-	if (item.m_count == 1)
+	// Draw damage amount
+	if (item.isDamaged()) {
+		int duraWidth = ceilf(13.0 - static_cast<double>(item.getDamageValue()) * 13.0 / static_cast<double>(item.getMaxDamage()));
```
BrentDaMage: Why is this double? Why not a float?

source/world/item/Item.hpp
```diff
@@ -43,19 +43,27 @@ class Item
 		SLOT_HEAD
 	};
 
+	enum ToolType
```
BrentDaMage: Old?

source/world/item/Tool.hpp
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+class Tool
+{
+public:
+    enum Type
```
BrentDaMage, line 6: Maybe move `Item::ToolType` here, call this TypeMask, then have Type just be a standard enum? Also, you'd want to replace `<< 0` with `<< PICKAXE`, or whatever.

BrentDaMage (changes requested):

source/world/tile/Tile.cpp
```diff
@@ -368,15 +368,15 @@ void Tile::initTiles()
...
+		->setToolTypes(Tool::PICKAXE, 2)
 		->setSoundType(Tile::SOUND_STONE)
 		->setDescriptionId("oreGold");
 
 	Tile::ironOre = (new OreTile(TILE_ORE_IRON, TEXTURE_ORE_IRON))
 		->init()
 		->setDestroyTime(3.0f)
 		->setExplodeable(5.0f)
-		->setToolType(Tool::PICKAXE, 1)
+		->setToolTypes(Tool::PICKAXE, 1)
```
BrentDaMage: Maybe call this `setToolTypesAndLevel`?

source/client/renderer/entity/ItemRenderer.cpp
```diff
@@ -193,8 +193,8 @@ void ItemRenderer::renderGuiItemOverlay(Font* font, Textures* textures, ItemStac
 
 	// Draw damage amount
 	if (item.isDamaged()) {
-		int duraWidth = ceilf(13.0 - static_cast<double>(item.getDamageValue()) * 13.0 / static_cast<double>(item.getMaxDamage()));
-		int duraPercent = ceilf(255.0 - static_cast<double>(item.getDamageValue()) * 255.0 / static_cast<double>(item.getMaxDamage()));
+		int duraWidth = ceilf(13.0 - static_cast<float>(item.getDamageValue()) * 13.0 / static_cast<float>(item.getMaxDamage()));
```
BrentDaMage: Add the f for these

source/world/level/Material.hpp
```diff
@@ -29,7 +29,7 @@ class Material
 	virtual bool isFlammable() const;
 	virtual Material* setNotAlwaysDestroyable();
 
-	Material* setToolTypes(unsigned int toolMask, int toolLevel = 0);
+	Material* setToolTypesAndLevel(unsigned int toolMask, int toolLevel = 0);
```
BrentDaMage, line 34: Can we make a version of this that requires the toolLevel param, then have a version that doesn't? Call one `setToolTypes` and one `setToolTypesAndLevel`. Then for the one with no tool level, instead of defaulting to 0 in the argument/function, we just don't set the level at all, which will use the default that's in the constructor.

BrentDaMage (changes requested):

BrentDaMage (approved):

PR #315, opened 2026-01-30
fix credits breaking when window is resized

Un1q32:


source/client/gui/screens/CreditsScreen.cpp
```diff
@@ -4,14 +4,6 @@
...
 CreditsScreen::CreditsScreen() :
 	m_btnBack(3, "Back")
-{
-}
-
-CreditsScreen::~CreditsScreen()
-{
-}
-
-void CreditsScreen::_initCreditsText()
```
BrentDaMage, line 14: Keep the credits text init in its own function, just call it in the constructor.

Un1q32: Why? The function would just be called from the constructor and serve no other purpose.

BrentDaMage: Because it keeps that single piece of functionality abstract, and makes it clear as to what its doing without the need for comments. It also makes the constructor look much cleaner and more concise.

BrentDaMage (approved):

PR #318, opened 2026-01-31
Fix android compiler warns

Un1q32:


platforms/android/main.cpp
```diff
@@ -34,12 +34,12 @@ int g_MousePosX, g_MousePosY;
 struct engine
 {
     struct android_app* androidApp;
-    EGLDisplay display = nullptr;
-    EGLSurface surface = nullptr;
-    EGLContext context = nullptr;
-    int animating = 0;
-    bool initted = false;
-    NinecraftApp* ninecraftApp = nullptr;
+    EGLDisplay display;
```
BrentDaMage, line 37: is not initing these fine?

Un1q32: the only time one is created is on line 401, and it gets memsetted to 0 there.

PR #327, opened 2026-02-01
Remove ELF file from NX artifacts

Bagietas:
.elf is inflating the Nintendo Switch artifact by 60mbs (actual size around 25mbs as compressed to zip), apparently reason was _Debug Symbols_ but who's gonna use pre-built binary for debugging?


Un1q32:
no

BrentDaMage:
> .elf is inflating the Nintendo Switch artifact by 60mbs (actual size around 25mbs as compressed to zip), apparently reason was _Debug Symbols_ but who's gonna use pre-built binary for debugging?

We need to have a version with symbols on-hand so we can utilize a stacktrace sent to us by an end-user.

PR #350, opened 2026-02-03
Sword and bow

samdotjpg:


source/world/item/BowItem.hpp
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+#include "Item.hpp"
+
+class BowItem : public Item
+{
+public:
+	BowItem(int id);
+
+	ItemStack* use(ItemStack* inst, Level* level, Player* player) const override;
```
samdotjpg: This should probably be "Mob" given the trajectory of other changes but I'm going to leave that to whoever wants to make the decision in changing that.

source/world/entity/Player.cpp
```diff
@@ -325,7 +325,33 @@ void Player::attack(Entity* pEnt)
 {
 	int atkDmg = m_pInventory->getAttackDamage(pEnt);
 	if (atkDmg > 0)
+	{
+		// Later beta did have a crit mechanic. It wasn't in b1.2_02 so it's commented out but feel free to re-enable it
```
BrentDaMage: Go ahead and uncomment this, we are targeting b1.7.3

game/assets/item/arrows.png
BrentDaMage, line 1: Where did this come from, and do we have the license to use it?

source/world/item/BowItem.hpp
BrentDaMage: You can go ahead and change it.

game/assets/item/arrows.png
samdotjpg: XSSheep's Pixel Perfection Legacy. I'll toss arrows.png it in the credits when I uncomment the rest, unless you want to use something else

BrentDaMage: Nah, that's fine.

source/world/item/CameraItem.cpp
```diff
@@ -18,27 +18,18 @@ CameraItem::CameraItem(int id) : Item(id)
...
-ItemStack* CameraItem::use(ItemStack* inst, Level* level, Player* player) const
+ItemStack* CameraItem::use(ItemStack* inst, Level* level, Mob* user) const
 {
 #ifndef ORIGINAL_CODE
 	// prevent players from using this in multiplayer, to prevent a desync of entity IDs
 	if (level->m_bIsClientSide)
 		return inst;
 #endif
 
-	/*Mob* entity = new Pig(level);
```
samdotjpg, line 29: This has to have been some debug code left over

BrentDaMage: We were using before we added the /summon command.

PR #352, opened 2026-02-03
Drop bush tile items when broken by an external source

samdotjpg:


BrentDaMage:
Grass is also a Bush as far as I know. Wouldn't this mean that breaking the block under grass would drop the grass? That doesn't seem right.

samdotjpg:
No, that doesn't happen. Grass isn't set to drop anything. Just the flowers and mushrooms

PR #371, opened 2026-02-04
Swords should be tools

samdotjpg:
Swords can break webs faster and to avoid hardcoding in a web check (which was there, _but wasn't even working anyway_, since the tiles check for compatibility with a tool), they really should just be a tool.

### DIFF
--- a/source/world/item/ToolItem.hpp
+++ b/source/world/item/ToolItem.hpp
@@ -9,16 +9,17 @@ class ToolItem : public Item
 public:
 	ToolItem(int id, Tool::Type toolType, Tier& tier);
 
+protected:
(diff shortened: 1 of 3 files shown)


source/world/item/ToolItem.hpp
```diff
@@ -16,8 +16,14 @@ class ToolItem : public Item
 	bool isHandEquipped() const override;
 	bool canDestroySpecial(const Tile* tile) const override;
 
+protected:
+	bool canDestroyTile(const Tile*) const;
```
BrentDaMage: Use underscore prefix for non-public functions. Also, put these above the public functions. It should go least-accessible to most-accessible.

Issue #374, opened 2026-02-05
Leaves don't decay unless a leaf block is broken


samdotjpg:
Does not match b1.2_02 or b1.7.3 behavior. When the final log is broken the leaves should start decaying immediately

https://github.com/user-attachments/assets/dbff6306-3e65-4a84-853c-4fa4697caae7

Wilyicaro:
I think the only difference is that it should be `data | 8` instead of `data | 4` in https://github.com/ReMinecraftPE/mcpe/blob/master/source/world/tile/TreeTile.cpp

samdotjpg:
Yeah it's 4 since b1.2_02 had it as 4 but by beta 1.7.3 it was 8 and the rest of the logic follows behind the idea it's 8. That does fix it

BrentDaMage:
> Yeah it's 4 since b1.2_02 had it as 4 but by beta 1.7.3 it was 8 and the rest of the logic follows behind the idea it's 8. That does fix it

Can you submit a PR?

Issue #373, opened 2026-02-05
ItemStack::isDamaged returns true for objects with an aux value > 0 that aren't damageable


samdotjpg:
```
// all items get max damage 32
Item::Item(int itemID)
{
	// ...
	m_maxDamage = 32;
	// ...
}

// returns 32 by default as a result
int Item::getMaxDamage() const
{
	return m_maxDamage;
}

// a "damageable item" is now any item with >m_maxDamage of 32 which means literally everything
bool ItemStack::isDamageableItem() const
{
	return getItem()->getMaxDamage() > 0;
}

// meaning this is basically "if true and my aux value is above 0"
bool ItemStack::isDamaged() const
{
	return isDamageableItem() && m_auxValue > 0;
}
```

BrentDaMage:
<img width="63" height="67" alt="Image" src="https://github.com/user-attachments/assets/78dcb2ab-c5fe-41d3-a6b1-61d18e08ef56" />

samdotjpg:
This is actually a bigger issue due to no implementation of item aux values for different item types (i.e. dyed powder). Giving them an aux value of say, 4, does not give you lapis lazuli, and instead just gives you damaged ink sacs lol

samdotjpg:
I'm going to fix all of this in my PR that has lapis ore drop lapis items.

Issue #372, opened 2026-02-04
Coal ore drops coal ore


samdotjpg:
Coal ore should just drop coal.

PR #385, opened 2026-02-05
Dyes, mob interactions

samdotjpg:
- dye colors and functionality implemented sans crop growing
- ores drop their correct items (addr. #372)
- items can set a max damage now - if an item is not supposed to mutate into other variants it **WILL** show a healthbar on it if you give it an aux value. this is the expected output from beta (addr. #373)
- trees decay properly now (addr. #374)

### DIFF
--- a/source/world/entity/Player.cpp
+++ b/source/world/entity/Player.cpp
@@ -525,7 +525,7 @@ void Player::interact(Entity* pEnt)
 		return;
 
 	ItemStack& item = getSelectedItem();
-	if (!item.isEmpty() && isMob) {
+	if (!item.isEmpty()) {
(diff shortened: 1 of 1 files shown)


source/world/entity/Player.cpp
```diff
@@ -517,7 +517,21 @@ void Player::touch(Entity* pEnt)
...
-	pEnt->interact(this);
+	if (pEnt->interact(this))
+		return;
+
+	bool isMob = pEnt->getDescriptor().hasCategory(EntityCategories::MOB);
+	if (!isMob)
+		return;
+
+	ItemStack& item = getSelectedItem();
+	if (!item.isEmpty() && isMob) {
```
BrentDaMage: No point in checking `isMob` here

PR #386, opened 2026-02-05
iOS incremental builds

Un1q32:
Also fixes a couple bugs:
LTO was always disabled.
If ldid build failed subsequent runs wouldn't try to build it.

### DIFF
--- a/.github/workflows/artifacts.yml
+++ b/.github/workflows/artifacts.yml
@@ -2,7 +2,7 @@ name: 'Build Artifacts'
 
 on:
   push:
-    branches: [ 'ios-incremental-builds' ]
+    branches: [ 'master' ]
 
 jobs:
     linux:


platforms/ios/build.sh
```diff
@@ -104,33 +100,33 @@ if [ -n "$outdated_toolchain" ]; then
...
+        rm -rf "ldid-$ldid_commit"
     fi
+    printf '%s' "$toolchainver" > "$workdir/bin/toolchainver"
 fi
 
 # checks if the linker we build successfully linked with LLVM and supports LTO,
 # and enables LTO in the cmake build if it does.
 if [ -z "$DEBUG" ]; then
-    if printf 'int main(void) {return 0;}' | "$target-cc" -xc - -flto -o "$workdir/testout" >/dev/null 2>&1; then
+    if printf 'int main(void) {return 0;}' | REMCPE_TARGET=armv7-apple-ios3.1 "$platformdir/ios-cc" -xc - -flto -o "$workdir/testout" >/dev/null 2>&1; then
```
BrentDaMage, line 124: I would prefer it if the target were a variable at the top of the script/

BrentDaMage (changes requested):

platforms/ios/build.sh
Un1q32: The target here doesn't matter at all, it just needs to build and link one binary to check if LTO works, it could be any arm iOS target, or even x86 macOS

Un1q32: Actually no it couldn't be macOS because it has to be a target the SDK supports, but it can be any target the SDK supports.

PR #399, opened 2026-02-06
Better handle lack of assets directory + native message function

Un1q32:


platforms/sdl/sdl2/base/AppPlatform_sdl2.cpp
```diff
@@ -192,3 +192,26 @@ bool AppPlatform_sdl2::GetMouseButtonState(const SDL_Event& event)
 
 	return result;
 }
+
+void AppPlatform_sdl2::showMessageModal(struct MessageModal msg)
```
BrentDaMage: nah, you don't have to specify `struct` here, also you're doing a copy, you should be doing `const MessageModal& msg`

platforms/sdl/sdl2/base/AppPlatform_sdl2.cpp
```diff
@@ -192,3 +192,26 @@ bool AppPlatform_sdl2::GetMouseButtonState(const SDL_Event& event)
...
+{
+	const char *title;
+	Uint32 flags = 0;
+
+	switch (msg.type)
+	{
+		case MessageModal::TYPE_ERROR:
+			title = "Error";
```
BrentDaMage: Actually, the title should absolutely be a part of the MessageModal. We should have the ctor not require a title, and have it set one by default based on the type, if the title is empty.

platforms/windows/AppPlatform_win32.cpp
```diff
@@ -570,3 +570,26 @@ Keyboard::KeyState AppPlatform_win32::GetKeyState(UINT iMsg)
...
+	uint32_t flags = 0;
+
+	switch (msg.type)
+	{
+		case MessageModal::TYPE_ERROR:
+			title = "Error";
+			flags = MB_ICONERROR;
+			break;
+		default:
+			LOG_W("Unhandled MessageModal type");
```
BrentDaMage: Just call the base function here, and have the base function log this.

source/client/gui/screens/CreditsScreen.cpp
```diff
@@ -32,6 +32,9 @@ void CreditsScreen::_initCreditsText()
 	{
 		m_credits.push_back("Failed to load credits.txt");
 	}
+
+	// for testing only, should not make it into master, remove if found
```
BrentDaMage: found it

Un1q32: nerd emoji

Un1q32: I'll remove it when I'm done, IE when the PR isn't a draft anymore.

platforms/sdl/sdl2/base/AppPlatform_sdl2.cpp
Un1q32: iOS and Android don't seem to have dedicated error and info messages, just a generic message, passing a title would mean theres no way to distinguish between the 2.

PR #409, opened 2026-02-07
Option Instances (and Gui Element improvements)

Wilyicaro:


game/assets/lang/en_us.json
BrentDaMage, line 1: I don't understand what's going on here. Why does this only contain options? Aren't the options already defined in another lang file??

source/client/gui/components/ScrolledSelectionList.cpp
```diff
@@ -160,80 +200,128 @@ void ScrolledSelectionList::render(const MenuPointer& pointer, float f)
...
+
+		t.begin(4);
+		t.color(0, 255);
+		t.vertexUV(barLeft, m_y1, 0.0, 0.0, 1.0);
+		t.vertexUV(barRight, m_y1, 0.0, 1.0, 1.0);
+		t.vertexUV(barRight, m_y0, 0.0, 1.0, 0.0);
+		t.vertexUV(barLeft, m_y0, 0.0, 0.0, 0.0);
+		t.draw(m_materials.ui_fill_gradient);
+		t.begin(4);
+		t.color(8421504, 255);
```
BrentDaMage: make this hex

source/client/gui/components/ScrolledSelectionList.cpp
```diff
@@ -160,80 +200,128 @@ void ScrolledSelectionList::render(const MenuPointer& pointer, float f)
...
+		int barTop = (int)m_scrollAmount * (m_y1 - m_y0 - barHeight) / yDiff + m_y0;
+		if (barTop < m_y0)
+			barTop = m_y0;
+
+		int barLeft = m_width / 2 + 124;
+		int barRight = barLeft + 6;
+
+		t.begin(4);
+		t.color(0, 255);
+		t.vertexUV(barLeft, m_y1, 0.0, 0.0, 1.0);
```
BrentDaMage: make these use `f`

source/client/gui/components/SmallButton.cpp
```diff
@@ -23,13 +23,26 @@ SmallButton::SmallButton(int id, int x, int y, int width, int height, const std:
...
-Options::Option* SmallButton::getOption()
+bool SmallButton::clicked(Minecraft* mc, int xPos, int yPos)
 {
-	return m_pOption;
+	if (!Button::clicked(mc, xPos, yPos)) return false;
+	if (getOption())
+	{
+		getOption()->toggle();
+		setMessage(getOption()->getMessage());
+		//mc->getOptions()->save();
```
BrentDaMage: remove this entirely

source/client/gui/components/TextBox.hpp
```diff
@@ -20,15 +20,14 @@ class Minecraft;
 
 // @TODO: Rename this to TextBox
```
BrentDaMage: remove the comment

source/client/gui/components/TextBox.hpp
```diff
@@ -20,15 +20,14 @@ class Minecraft;
...
 // Don't trash this in favor of Mojang's class, just mold it, since this is better at handling all platforms
```
BrentDaMage: and this ig

source/client/options/Options.cpp
```diff
@@ -21,82 +21,105 @@
...
+	m_thirdPerson.set(0);
 	field_19 = 0;
 #endif
 
 	loadControls();
 }
 
-Options::Options()
+Options::Options(Minecraft* mc, const std::string& folderPath) :
+	m_pMinecraft(mc),
```
BrentDaMage: Please format this with the colons and commas on the left side, that way git does a better job at merging. This is also the convention Mojang uses.

source/client/options/Options.cpp
```diff
@@ -235,6 +223,13 @@ std::string Options::saveInt(int i)
 	return ss.str();
 }
 
+std::string Options::saveFloat(float f)
```
BrentDaMage, line 235: we should really have this get passed a stringstream reference, that way all saveX functions use the same stringstream.

source/client/options/Options.cpp
```diff
@@ -574,3 +555,80 @@ void Options::loadControls()
...
+	if (!GrassColor::isAvailable() && !FoliageColor::isAvailable())
+		m_biomeColors.set(false);
+
+	if (!Screen::isMenuPanoramaAvailable())
+		m_menuPanorama.set(false);
+}
+
+const std::string& Options::OptionEntry::getName() const
+{
+	return Language::singleton().get(m_name);
```
BrentDaMage: imo, `get` should be a static method of `Language`, which just does `Language::singleton().get(name)`. that'd make all the calls cleaner.

game/assets/lang/en_us.json
Wilyicaro: The plan, as in the comment in `Language.cpp`, is to maintain intact the original `en_US.lang` and make modifications only in the `en_us.json`

source/client/gui/components/SmallButton.cpp
Wilyicaro: The comment?

source/client/gui/components/ScrolledSelectionList.cpp
Wilyicaro: The other renderings in this file also use decimals, so I'll change them too

source/client/options/Options.cpp
Wilyicaro: I agree

game/assets/lang/en_us.json
BrentDaMage: Sounds good

source/client/gui/components/SmallButton.cpp
BrentDaMage: yes. there is no reality where we want to do an I/O operation every time an option is toggled

source/client/options/Options.cpp
BrentDaMage: btw, if u don't wanna do this, just put an `@TODO` comment

Wilyicaro: I did this, but I'm not using this function anymore

BrentDaMage: If the function isn't used, it should probably be deleted.

BrentDaMage: Upon reviewing these functions and their use case, I think it would be more appropriate to have this in some sort of  separate static `PlainTextSerializer` class. But for now, I don't think it makes any sense to hold this PR up any further.

Wilyicaro: I think the same

PR #414, opened 2026-02-07
Touchscreen hearts rendering

samdotjpg:
Stop the touchscreen progress indicator from messing up the shading of the hearts in the UI.


BrentDaMage:
Why does this change stuff that's already been pulled into master?

PR #416, opened 2026-02-08
GLES lighting fix

samdotjpg:
Very complicated PR, please take the time to read every line.

### DIFF
--- a/source/renderer/hal/ogl/RenderContextOGL.cpp
+++ b/source/renderer/hal/ogl/RenderContextOGL.cpp
@@ -148,8 +148,8 @@ void RenderContextOGL::enableFixedLighting(bool init)
 #ifdef USE_GL_NORMAL_LIGHTING
         glEnable(GL_LIGHT0);
         glEnable(GL_LIGHT1);
-#if !defined(__EMSCRIPTEN__) && !defined(USE_GLES)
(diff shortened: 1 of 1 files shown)


source/renderer/hal/ogl/RenderContextOGL.cpp
```diff
@@ -148,8 +148,8 @@ void RenderContextOGL::enableFixedLighting(bool init)
 #ifdef USE_GL_NORMAL_LIGHTING
         glEnable(GL_LIGHT0);
         glEnable(GL_LIGHT1);
-#if !defined(__EMSCRIPTEN__) && !defined(USE_GLES)
         glEnable(GL_COLOR_MATERIAL);
```
BrentDaMage, line 151: this doesn't exist in GLES 2, nor does any of this for that matter. but as long as it builds, fuck it

BrentDaMage (approved):

source/renderer/hal/ogl/RenderContextOGL.cpp
samdotjpg: Hang on

BrentDaMage: my bad, didn't see this

PR #419, opened 2026-02-08
Better handle lack of assets directory + native message function

Un1q32:
reopening this cuz i made it its own branch

### DIFF
--- a/platforms/ios/AppPlatform_iOS.mm
+++ b/platforms/ios/AppPlatform_iOS.mm
@@ -194,6 +194,7 @@ void AppPlatform_iOS::showMessageModal(const MessageModal& msg)
 			break;
 	}
 
+	// this doesn't block the thread, it should
 	UIAlertView *alert = [[UIAlertView alloc]
 		initWithTitle:title
 		message:[NSString stringWithCString:msg.text.c_str()


platforms/sdl/sdl2/base/AppPlatform_sdl2.cpp
```diff
@@ -192,3 +192,27 @@ bool AppPlatform_sdl2::GetMouseButtonState(const SDL_Event& event)
 
 	return result;
 }
+
+// this segfaults on wsl, why?
```
BrentDaMage, line 196: how the hell is SDL supposed to show a dialog box in WSL?

Un1q32: By using GTK or QT or something idk.  It shouldn't *crash* it should at least return some kind of fail state so we can fallback to stderr.

BrentDaMage: true

BrentDaMage:
Is there a reason the branch was deleted? I feel like someone could've worked off of this.

Un1q32:
The diff is still archived by GitHub.  I might restart it with a much smaller scope, only implementing the message box for windows and letting everyone else use printf to stderr.  I never finished it cuz I couldn't figure out how to make every platform behave consistently, some platforms block the thread until the box is dismissed, some don't, some have titles, some don't.  It was very messy, it would be a lot easier and simpler to just *only* handle the text body and set a filler title.

PR #421, opened 2026-02-08
Flying

### LINKED ISSUE #398: Normal Flight Controls + Keep Noclip Flight
Flying should be like in JE creative mode, double tap jump to start flying, double tap again to stop, go up with jump and down with sneak.
Noclipping should be only in a dedicated spectator mode gamemode that behaves like JE spectator mode.

### LINKED ISSUE #401: Every wool color in the creative inventory points to red wool
Need to query both the block and its subtype.

Yellow wool selected:
Image

### LINKED ISSUE #410: The crafting text in the inventory renders above items
Image

samdotjpg:
Flying with KBM/controller. You can fly in creative and you get booted out in survival mode. Intentionally does not change FOV by 1.1x to match Beta's hand/screen FOV matching.

Closes #398, fixes #401 (checks if item has subtypes and if it matches), and fixes #410

### DIFF
--- a/source/world/entity/Player.hpp
+++ b/source/world/entity/Player.hpp
@@ -49,6 +49,7 @@ public:
 	bool isShootable() const override { return true; }
 	bool isPlayer() const override { return true; }
 	bool isCreativeModeAllowed() const override { return true; }
+	bool isSlowedByLiquids() const override { return !m_bFlying; }
 	bool hurt(Entity*, int) override;
 	void awardKillScore(Entity* pKilled, int score) override;
 	void resetPos(bool respawn = false) override;
@@ -62,7 +63,6 @@ public:
 	void readAdditionalSaveData(const CompoundTag& tag) override;
 	void travel(const Vec2& pos) override;
 	void causeFallDamage(float level) override;
-	bool isSlowedByLiquids() const override { return !m_bFlying; }
 
 	virtual void animateRespawn();
 	//virtual void drop(); // see definition


Wilyicaro:
This will cause issues with the touchscreen controls, as the jumping is endless

Change the line 190 of `TouchscreenInput_TestFPS.cpp` to fix i:
`else if (Multitouch::isPressed(finger) || pPlayer->m_bFlying)`

samdotjpg:
Thanks for checking that for me!

source/client/player/LocalPlayer.cpp
```diff
@@ -67,17 +67,48 @@ void LocalPlayer::die(Entity* pCulprit)
...
+
 	if (m_pMoveInput->m_bSneaking && m_ySlideOffset < 0.2f)
 		m_ySlideOffset = 0.2f;
 
 	m_lastRenderArmRot = m_renderArmRot;
 	m_renderArmRot.x = Mth::Lerp(m_renderArmRot.x, m_rot.x, 0.5f);
 	m_renderArmRot.y = Mth::Lerp(m_renderArmRot.y, m_rot.y, 0.5f);
 
+	// timer for switching flight states
+	if (isCreative() && m_pMoveInput->m_bJumping && !wasJumping)
```
BrentDaMage: imo, we shouldn't be checking if the player is in creative, but rather, whether or not flight is enabled. Mods may want to give the player flight without putting them into creative mode.

source/world/entity/Player.cpp
```diff
@@ -307,6 +311,33 @@ void Player::readAdditionalSaveData(const CompoundTag& tag)
...
+	Mob::travel(pos);
+	
+	m_flyingFriction = oldFlyingFriction;
+	m_vel.y = yd * 0.6;
+}
+
+void Player::causeFallDamage(float level)
+{
+	// There is absolutely no reason for this to be causing the bone cracking sound in creative mode.
+	if (!isCreative())
```
BrentDaMage: If possible, this should just check if the player is invulnerable.

source/world/entity/Player.hpp
```diff
@@ -19,6 +19,13 @@ class Inventory; // in case we're included from Inventory.hpp
 
 class Player : public Mob
 {
+public:
```
samdotjpg, line 22: I had this as public since I had an extra ability I ended up removing. Can change this to protected (or private, if you really want the overhead for grabbing the reference from a getter function)

source/world/entity/Player.hpp
```diff
@@ -19,6 +19,13 @@ class Inventory; // in case we're included from Inventory.hpp
...
+	struct Abilities
```
BrentDaMage, line 23: Since this is a struct, you can remove the m_ prefix.

source/world/entity/Player.hpp
BrentDaMage: To remove the overhead of a getter function, you can just inline it. Please make it protected. Thanks.

samdotjpg:
It won't let me respond directly to your comment- do you want the getter function? And by make it protected, do you mean the getter function or the Abilities struct itself? If the struct is protected, you can't reference it with Player::Abilities& since it would be a protected struct. So I assume what you mean is no getter function, and make the abilities protected then?

BrentDaMage:
> It won't let me respond directly to your comment- do you want the getter function? And by make it protected, do you mean the getter function or the Abilities struct itself? If the struct is protected, you can't reference it with Player::Abilities& since it would be a protected struct. So I assume what you mean is no getter function, and make the abilities protected then?

Do the getter function, so it can be accessed outside of the player, and make the field protected.

source/world/entity/Player.hpp
samdotjpg: Okay, the struct will still need to remain public for the getter to work and be assignable to a Player::Abilities&.

source/world/entity/Player.hpp
```diff
@@ -110,6 +111,9 @@ class Player : public Mob
 	void interact(Entity* pEnt);
 #pragma GCC diagnostic pop
 
+protected:
+	Abilities abilities;
```
BrentDaMage: no no no, I meant you can remove the `m_` prefix for the struct *members*, not for the variable the struct is stored in

source/world/entity/Player.hpp
```diff
@@ -94,7 +94,7 @@ class Player : public Mob
 	void rideTick();
 	void setDefaultHeadHeight();
 	void setRespawnPos(const TilePos& pos);
-	inline Abilities& getAbilities() { return abilities; }
+	inline Abilities& getAbilities() { return m_abilities; }
```
BrentDaMage: this should return a const reference, and the function should be const

PR #435, opened 2026-02-09
Mob Spawning

### LINKED ISSUE #425: Autojump is broken
Attempting to auto jump while facing straight and moving on the Z axis doesn't work at all, -/+. Works fine if there's an amount of X momentum (camera tilt to the side or move diagonally). Works fine on X axis.

### LINKED ISSUE #426: Doors only give you an item if you break the bottom tile
They do not give you an item if you break the top tile or break the tile it's standing on.

samdotjpg:
Directly based off #237's implementation (thank you @rep-stosd) but I didn't want to fork the fork and update the fork and hm forks. Anyway, this is not biome-specific mob spawning, this is alpha-style mob spawning. I would like to also add biome mob spawning at some point but this is fine for now I think.

Specific nuances have been addressed from comments:
- all/allCount name changes to MobCategory/EntityCategories
- Add ampersand within comment. (MobCategory::all)
- Might be better as a map or a switch statement, just a suggestion. / You don't need to be fetching the mask manually, you can just compare the EntitityCategories to a mask directly. (MobFactory::GetMobListOfCategory)
- Don't bother fetching the CategoriesMask directly, just compare the EntityCategories to a CategoriesMask. (on Level::entityAdded/Level::entityRemoved)
- Make this return an unsigned int. (Level::getEntityCount)

And in doing so, I have, of course, likely introduced plenty of other things that will now need to be addressed.

Also fixes #425, fixes #426, fixes segfault when opening chat with /


source/world/entity/MobFactory.cpp
```diff
@@ -24,6 +24,13 @@
 
 #define ENT(enumType, classType) case EntityType::enumType: return new classType(level);
 
+// format: ID, spawnrate
+std::map<EntityType::ID, int> monsterList;
+std::map<EntityType::ID, int> creatureList;
+std::map<EntityType::ID, int> waterCreatureList;
+std::map<EntityType::ID, int> nullCreatureList;
+std::map<EntityCategories::CategoriesMask, std::map<EntityType::ID, int>*> mobListsByCategory;
```
samdotjpg, line 32: Maybe aggregate this so that it contains monster/creature/water inside of it instead of by pointer.

source/world/entity/MobFactory.cpp
```diff
@@ -35,4 +42,34 @@ Mob* MobFactory::CreateMob(EntityType::ID entityType, Level *level)
     }
 }
 
-#undef ENT
+void MobFactory::initMobLists() 
+{
+    // format: ID, spawnrate
+    mobListsByCategory.insert(std::make_pair(EntityCategories::MONSTER, &monsterList));
```
samdotjpg, line 48: Comment on mobListsByCategory may address the "ugliness" of this. Asking the council to advise

source/world/entity/MobSpawner.hpp
```diff
@@ -0,0 +1,30 @@
...
+#include "world/level/levelgen/chunk/LevelChunk.hpp"
+#include "world/level/levelgen/chunk/ChunkSource.hpp"
+#include "world/level/storage/LevelStorageSource.hpp"
+#include "world/level/storage/LevelSource.hpp"
+#include "world/level/storage/LevelData.hpp"
+#include "world/level/path/PathFinder.hpp"
+
+class MobSpawner {
+private:
+    static bool isSpawnPositionOk(const MobCategory& category, Level& level, const TilePos& pos);
```
samdotjpg: Saw conflicting files on whether to capitalize static methods. Let me know which way is the truly preferred choice and this (likely) won't come up again

BrentDaMage: Yes, please capitalize static methods. The only scenarios where this isn't done is when something is directly decompiled, but even then, we should probably still be doing it.

source/world/entity/MobSpawner.cpp
```diff
@@ -0,0 +1,197 @@
...
+TilePos MobSpawner::getRandomPosWithin(Level& level, int chunkX, int chunkZ) 
+{
+    int px = level.m_random.nextInt(16) + chunkX;
+    int py = level.m_random.nextInt(128);
+    int pz = level.m_random.nextInt(16) + chunkZ;
+    return TilePos(px, py, pz);
+}
+
+//todo: bool?
+int MobSpawner::addMob(Level& level, Mob *mob, const Vec3& pos, const Vec2& rot)
```
BrentDaMage: Make this a bool for now, if we need to change it to an enum, we can do that later.

source/world/level/Level.cpp
```diff
@@ -316,6 +319,16 @@ Entity* Level::getEntity(Entity::ID id) const
 	return nullptr;
 }
 
+unsigned int Level::getEntityCount(const EntityCategories& category) const
+{
+	// @TODO: probably change this to a BST at some point. this works for now though
```
BrentDaMage: put the BST comment in entityAdded

BrentDaMage (changes requested):
A couple minor changes.

source/world/entity/MobFactory.cpp
BrentDaMage: fine enough

source/world/entity/MobFactory.cpp
BrentDaMage: Nah, I mean this is probably fine. If you're concerned about it, definitely put a mention of it in a comment somewhere.

BrentDaMage (approved):

PR #437, opened 2026-02-09
Touch button additions and improvements

Vimd0:
Screenshot (623)

https://github.com/user-attachments/assets/58403dd1-fdf7-475b-b397-541bea066b7c

* New touch control textures and ui elements (created by me)
* Hotbar now supports a 10th slot (Keeping at 6 for touch screens for now until we have a get screen width feature to set the number of slots.)
* Reworked the options menu a bit
* Added pack.png

### DIFF
--- a/source/client/gui/components/WorldSelectionList.cpp
+++ b/source/client/gui/components/WorldSelectionList.cpp
@@ -188,7 +188,7 @@ void WorldSelectionList::commit()
 		/*std::stringstream ss;
 		ss << item.m_levelName << "/preview.png";*/
 
-		m_previewImages.push_back("/pack.png");
+		m_previewImages.push_back("pack.png");
 
 		std::vector<std::string> vs;
 		vs.push_back(item.m_levelName);


Un1q32:
put the new dpad in gui_touch.png instead so beta 1.7.3 resource packs that override gui.png won't break things.

Vimd0:
> put the new dpad in gui_touch.png instead so beta 1.7.3 resource packs that override gui.png won't break things.

oh shi you right

Un1q32:
Also credit yourself for the texture in game/assets/readme.txt

Un1q32:
Replace "parts of gui_custom.png" with just "gui_custom.png".  It looks like everything in there is by you.
Also remove gui_touch.png since it's unused now.

Vimd0:
> Replace "parts of gui_custom.png" with just "gui_custom.png". It looks like everything in there is by you. Also remove gui_touch.png since it's unused now.

already removed

Un1q32:
Sorry im blind

Un1q32:
Is there a change in gui.png im not seeing or is it just reencoded?
It's 1.6k smaller so if it is just keep the new one.

Vimd0:
> Is there a change in gui.png im not seeing or is it just reencoded? It's 1.6k smaller so if it is just keep the new one.

I removed the cross hair at the top right. We don't need it since it draws in game.

source/client/player/input/TouchscreenInput_TestFps.cpp
```diff
@@ -336,35 +336,46 @@ static void RenderTouchButton(Tesselator* t, PolygonArea* pArea, int srcX, int s
...
 	t.begin(0);
 #ifdef ENH_NEW_TOUCH_CONTROLS
 	if (m_bForwardBeingHeld && !isButtonDown(100 + INPUT_JUMP)) 
 	{
-		t.color(isButtonDown(100 + INPUT_FORWARDLEFT) ? 0xC0C0C0 : 0xFFFFFF, 0x80);
+		t.color(isButtonDown(100 + INPUT_FORWARDLEFT) ? 0xFFFFFF : 0xFFFFFF, 0x80);
 		RenderTouchButton(&t, m_pAreaForwardLeft, 0, 132);
 		
-		t.color(isButtonDown(100 + INPUT_FORWARDRIGHT) ? 0xC0C0C0 : 0xFFFFFF, 0x80);
+		t.color(isButtonDown(100 + INPUT_FORWARDRIGHT) ? 0xFFFFFF : 0xFFFFFF, 0x80);
```
samdotjpg: Why are we performing this check if it's the same output for both conditions?

GameMods.hpp
```diff
@@ -41,15 +41,14 @@
 #define ENH_3D_INVENTORY_TILES  	       // Uses 3D rendered inventory tiles, use with ENH_SHADE_HELD_TILES to render correctly.
 #define ENH_MENU_BACKGROUND 		       // Renders a spinning panorama (if it's available) in the background of the main menu
 #define ENH_GUI_ITEM_POP 	               // Calls Inventory::tick() to create the "pop" animation for items that enter the hotbar. This function was not present on Pocket Edition.
+#define ENH_NEW_TOUCH_CONTROLS 	           // Use the 0.6.0 touch controls
```
samdotjpg, line 44: 0.6.0 controls on by default?

source/client/player/input/TouchscreenInput_TestFps.cpp
```diff
@@ -241,14 +241,14 @@ void TouchscreenInput_TestFps::tick(Player* pPlayer)
 				break;
 
 			case 100 + INPUT_FORWARDLEFT:
-				if (m_bForwardBeingHeld) {
```
samdotjpg, line 244: If these conditionals are being removed, the brackets surrounding them should be as well, as per other switch statements.

game/assets/gui/gui_custom.png
BrentDaMage, line 1: Wait, why is the chat button in a different location, and why are the variants of the chat and pause buttons missing?

source/client/player/input/TouchscreenInput_TestFps.cpp
```diff
@@ -336,35 +336,46 @@ static void RenderTouchButton(Tesselator* t, PolygonArea* pArea, int srcX, int s
 
 void TouchscreenInput_TestFps::render(float f)
 {
-	std::string gui = "gui/gui_touch.png";
+	std::string gui = "gui/custom.png";
```
BrentDaMage: What is this?

BrentDaMage (changes requested):
Sam mentioned all the other points of concern.

Un1q32:
Maybe we should go back to having the dpad in gui_touch.png.  The idea was if you wanted to use mojang assets all you would have to do was copy mojang's gui.png to gui_touch.png and it would work.  Using gui_custom.png for this breaks that.

source/client/player/input/TouchscreenInput_TestFps.cpp
Vimd0: Doing this to prevent a syntax error

Vimd0: Doing this to prevent a syntax error

GameMods.hpp
Vimd0: yes

source/client/player/input/TouchscreenInput_TestFps.cpp
Vimd0: Accidentally removed these

game/assets/gui/gui_custom.png
Vimd0: The chat variants were removed for resource pack reasons, and I repositioned the elements to match the PE gui

source/client/player/input/TouchscreenInput_TestFps.cpp
Vimd0: Oops. I fixed it

Vimd0:
> Maybe we should go back to having the dpad in gui_touch.png. The idea was if you wanted to use mojang assets all you would have to do was copy mojang's gui.png to gui_touch.png and it would work. Using gui_custom.png for this breaks that.

I repositioned the items to match PEs, and I'm doing this because the assets are becoming way too cluttered. gui_custom wasn't being used to its full potential

source/client/player/input/TouchscreenInput_TestFps.cpp
BrentDaMage: How does doing this prevent a syntax error?

game/assets/gui/gui_custom.png
BrentDaMage: Okay wait, these buttons have the same shared purpose between ReMCPE and MCPE, why would we not put them were MCPE puts them, that way an MCPE resource pack would override them correctly? Also, do we just not need the variants that are used when the button is pressed??

source/client/gui/components/OptionList.cpp
```diff
@@ -150,23 +150,26 @@ void OptionList::initDefaultMenu()
...
-void OptionList::initVideoMenu()
+void OptionList::initGameplayMenu()
 {
 	Options* pOptions = m_pMinecraft->getOptions();
-	int currentIndex = 0;
-	(void)currentIndex;
+	int currentIndex = -1;
 
-	OPTION(m_viewDistance);
+	HEADER(Util::EMPTY_STRING);
```
BrentDaMage: This is useless padding, no?

source/client/gui/components/OptionList.cpp
```diff
@@ -175,44 +178,53 @@ void OptionList::initControlsMenu()
 	int currentIndex = -1;
 	int idxSplit = -1, idxController = -1;
 
-	OPTION(m_autoJump);
+	HEADER(Util::EMPTY_STRING);
```
BrentDaMage: This is useless padding, no?

source/client/gui/components/OptionList.cpp
```diff
@@ -175,44 +178,53 @@ void OptionList::initControlsMenu()
...
+void OptionList::initVideoMenu()
 {
 	Options* pOptions = m_pMinecraft->getOptions();
 	int currentIndex = -1;
-
-	OPTION(m_difficulty);
-	OPTION(m_serverVisibleDefault);
+	int idxPano = -1;
 
 	HEADER(Util::EMPTY_STRING);
```
BrentDaMage: This is useless padding, no?

source/client/gui/components/WorldSelectionList.cpp
```diff
@@ -188,7 +188,7 @@ void WorldSelectionList::commit()
 		/*std::stringstream ss;
 		ss << item.m_levelName << "/preview.png";*/
 
-		m_previewImages.push_back("gui/default_world.png");
+		m_previewImages.push_back("/pack.png");
```
BrentDaMage: Isn't default_world.png a different size?

source/client/gui/screens/OptionsScreen.cpp
```diff
@@ -13,16 +13,15 @@
...
 #define BACK_BUTTON_ID 100
 
 OptionsScreen::OptionsScreen() :
 	m_pList(nullptr),
 	m_currentCategory(OC_MIN),
-	m_videoButton(MIN_CATEGORY_BUTTON_ID, "Video"),
-	m_controlsButton(MIN_CATEGORY_BUTTON_ID + 1, "Controls"),
-	m_gameplayButton(MIN_CATEGORY_BUTTON_ID + 2, "Gameplay"),
-	m_miscButton(MAX_CATEGORY_BUTTON_ID, "Misc"),
+	m_gameplayButton(MIN_CATEGORY_BUTTON_ID, "Gameplay..."),
```
BrentDaMage: Wouldn't this make it fit worse on smaller screens?

source/client/player/input/TouchscreenInput_TestFps.cpp
```diff
@@ -336,35 +338,46 @@ static void RenderTouchButton(Tesselator* t, PolygonArea* pArea, int srcX, int s
 
 void TouchscreenInput_TestFps::render(float f)
 {
-	std::string gui = "gui/gui_touch.png";
+	std::string gui = "gui/gui_custom.png";
```
BrentDaMage: Dude, this whole if check fallback thing is doing nothing. You're checking the same file path twice.

BrentDaMage:
> > Maybe we should go back to having the dpad in gui_touch.png. The idea was if you wanted to use mojang assets all you would have to do was copy mojang's gui.png to gui_touch.png and it would work. Using gui_custom.png for this breaks that.
> 
> I repositioned the items to match PEs, and I'm doing this because the assets are becoming way too cluttered. gui_custom wasn't being used to its full potential

The assets should be set up in a way that allows the user to stick the vanilla Minecraft assets in a resource pack, and have it apply to as much as possible correctly. We should not be referencing the d-pad texture in gui_custom.png, instead it should be in gui_touch.png or in gui.png, where actual PE expects it to be.

source/client/gui/components/OptionList.cpp
Vimd0: You're right. Removed

source/client/gui/components/OptionList.cpp
Vimd0: You're right. Removed

source/client/gui/components/OptionList.cpp
Vimd0: You're right. Removed

source/client/gui/screens/OptionsScreen.cpp
Vimd0: Screenshot 2026-02-10 211840
This would do the complete opposite, because we had four buttons and would go off screen.

BrentDaMage: no, I mean wouldn't adding the "..." make it fit worse on screen?

source/client/player/input/TouchscreenInput_TestFps.cpp
Vimd0: This string stuff was already present before I made this pr. I removed it.

source/client/gui/components/WorldSelectionList.cpp
Vimd0: Screenshot (625)

Yeah, but it doesn't change much. Uniq added the texture.

BrentDaMage: ok, whatever. at some point we will have the world preview feature working.

source/client/player/input/TouchscreenInput_TestFps.cpp
Vimd0: We need these checks because they determine how the buttons appear.  (pressed color)0xFFFFFF:(idle color)0xFFFFFF, (opacity)0x80);. I made it always white for PE parity

BrentDaMage: Right, but we're wondering why you have the isButtonDown check if you're always returning the same value?

source/client/gui/screens/OptionsScreen.cpp
Wilyicaro: I don't think the "..." makes sense for tabs

game/assets/gui/gui_custom.png
Vimd0: Gui2 has been added for PE-related textures. Users can rename the PE GUI to "gui2". I also changed the texture for the on and off switch in gui_custom because the old one looked inconsistent with our new textures. This one will look better overall, regardless of the texture pack we use. The variant textures appeared to be an idea from iProgramCpp's early development for ReMCPE, but were never used.

Vimd0:
> > > Maybe we should go back to having the dpad in gui_touch.png. The idea was if you wanted to use mojang assets all you would have to do was copy mojang's gui.png to gui_touch.png and it would work. Using gui_custom.png for this breaks that.
> > 
> > 
> > I repositioned the items to match PEs, and I'm doing this because the assets are becoming way too cluttered. gui_custom wasn't being used to its full potential
> 
> The assets should be set up in a way that allows the user to stick the vanilla Minecraft assets in a resource pack, and have it apply to as much as possible correctly. We should not be referencing the d-pad texture in gui_custom.png, instead it should be in gui_touch.png or in gui.png, where actual PE expects it to be.

read my response in https://github.com/ReMinecraftPE/mcpe/pull/437#pullrequestreview-3775965969

game/assets/gui/gui_custom.png
BrentDaMage: Okay, to confirm, I would like our texture atlas for GUI textures to match PE & Java as much as possible. What you're saying is that the button variants in our gui_custom.png do not actually exist in PE, and that the new positions you put the buttons in *are* accurate to PE?

source/client/player/input/TouchscreenInput_TestFps.cpp
Vimd0: I assumed you two were discussing the colors lol. The other isButtonDowns are obviously required, but we need the isButtonDown check to ensure that they still render when the jump button is used. I also removed all of the unnecessary m_bForwardBeingHeld checks when moving forward, because Uniq was correct about it looking weird.

game/assets/gui/gui_custom.png
Vimd0: Exactly what I intended

source/client/gui/screens/OptionsScreen.cpp
Vimd0: Removed them

game/assets/gui/gui_custom.png
BrentDaMage: > Exactly what I intended

Great

BrentDaMage, line 1: What is this texture actually for? Are we actually using those switch textures?

Un1q32:
> instead it should be in gui_touch.png or in gui.png, where actual PE expects it to be.

Not in gui.png, that would make beta 1.7.3 packs break the touch stuff and we want to be able to use beta 1.7.3 resource packs as-is.

game/assets/gui/gui_custom.png
Vimd0: We're using these yeah

Vimd0:
> > instead it should be in gui_touch.png or in gui.png, where actual PE expects it to be.
> 
> Not in gui.png, that would make beta 1.7.3 packs break the touch stuff and we want to be able to use beta 1.7.3 resource packs as-is.

Screenshot_20260211_124811_My Files

I separated them from each other. We can easily swap gui2 with the pe gui, without any conflicts

game/assets/gui/gui_custom.png
BrentDaMage: first, why not just add these switch textures to gui2.png?
second, they look like abstract placeholders, could you please draw something else? I will give it a go if need be.

Vimd0: added touchgui

source/client/gui/components/OptionList.cpp
```diff
@@ -175,44 +176,49 @@ void OptionList::initControlsMenu()
...
-
-void OptionList::initMiscMenu()
+void OptionList::initVideoMenu()
 {
 	Options* pOptions = m_pMinecraft->getOptions();
 	int currentIndex = -1;
 	int idxPano = -1;
 
+	HEADER("Graphics");
+	//OPTION(m_brigtness);
```
BrentDaMage: brigtness

source/client/gui/components/OptionList.cpp
```diff
@@ -175,44 +176,49 @@ void OptionList::initControlsMenu()
...
+	OPTION(m_viewDistance);
+	//OPTION(m_antiAliasing);
+	OPTION(m_guiScale);
```
BrentDaMage: changing this crashes the game

Vimd0: commented out

source/client/gui/components/OptionList.cpp
Vimd0: fixed

BrentDaMage: it doesn't look like it's fixed

Vimd0: Now it's fixed

source/client/gui/components/WorldSelectionList.cpp
BrentDaMage: Please remove the forward slash at the beginning here, I think it may break support on the 360.

Vimd0: ok removed

PR #438, opened 2026-02-09
Food

samdotjpg:
It's Beta food- right click to eat.


source/world/item/FoodItem.cpp
```diff
@@ -0,0 +1,17 @@
+#include "FoodItem.hpp"
+#include "world/entity/Player.hpp"
+
+FoodItem::FoodItem(int id, int nutrition) : Item(id), m_nutrition(nutrition)
+{
+	m_maxStackSize = 1;
```
samdotjpg, line 6: I don't know, it works fine at a larger stack size. My one major gripe about Beta was food clogging the inventory. I had it at 8 but for now it's just adhering to what it should be

source/world/item/FoodItem.cpp
```diff
@@ -0,0 +1,17 @@
...
+}
+
+ItemStack* FoodItem::use(ItemStack* inst, Level* level, Mob* mob) const
+{
+    if (mob->m_health < 20)
```
BrentDaMage: We should be comparing this against the consumer's max health, not just 20.

BrentDaMage (changes requested):

source/world/entity/Player.cpp
```diff
@@ -178,7 +178,7 @@ void Player::aiStep()
 		m_jumpTriggerTime--;
 
     if (m_pLevel->m_difficulty == 0 &&
```
BrentDaMage, line 180: We do natural regen on peaceful? Is this a part of beta?

BrentDaMage (approved):

PR #439, opened 2026-02-10
macOS universal artifact part 2; i386 Tiger

Un1q32:
Adds an i386 slice that supports Tiger, and runs the Tiger version on x86_64 systems running snow leopard and below.

### DIFF
--- a/platforms/ios/build.sh
+++ b/platforms/ios/build.sh
@@ -165,6 +165,7 @@ for target in $targets; do
         -DCMAKE_BUILD_TYPE="$build" \
         -DCMAKE_SYSTEM_NAME=Darwin \
         -DREMCPE_PLATFORM=ios \
+        -DCMAKE_FIND_ROOT_PATH_MODE_PACKAGE=ONLY \
         -DCMAKE_FIND_ROOT_PATH_MODE_LIBRARY=ONLY \
         -DCMAKE_FIND_ROOT_PATH_MODE_INCLUDE=ONLY \
         -DCMAKE_AR="$(command -v "$ar")" \


BrentDaMage:
Why are we using SDL1.2 for x86_64?

Un1q32:
> Why are we using SDL1.2 for x86_64?

So we can support Snow Leopard, the version of SDL2 we've been using won't build for anything older than Lion.

BrentDaMage:
> > Why are we using SDL1.2 for x86_64?
> 
> So we can support Snow Leopard, the version of SDL2 we've been using won't build for anything older than Lion.

But I've been able to use the SDL2 version from MacPorts on Leopard and above without issues.

Un1q32:
> > > Why are we using SDL1.2 for x86_64?
> > 
> > 
> > So we can support Snow Leopard, the version of SDL2 we've been using won't build for anything older than Lion.
> 
> But I've been able to use the SDL2 version from MacPorts on Leopard and above without issues.

They're prolly using an older version of SDL 2 or they patch it.  You could make the script also manually build an older SDL2 but that would be annoying.  If the SDL 1.2 version is fixed to not have any issues than wouldn't it be better to use that if we can if for no other reason than that it makes the binary size smaller.

BrentDaMage:
No, because we cannot capture the scroll wheel on SDL1, and losing the render context every time the window resized is just a waste of CPU and IO. We want the best experience for the user as possible. Additionally, SDL1 is not capable of using the system's cursor, and will display its own cursor in the window, this is somewhat odd, but not a huge deal.

platforms/ios/build.sh
```diff
@@ -27,8 +27,9 @@ if ! [ -d "$sdk" ] || [ "$(cat sdkver 2>/dev/null)" != "$sdkver" ]; then
     # It also doesn't use tbd stubs so we don't need to link ld64 with libtapi.
     printf '\nDownloading iOS SDK...\n\n'
     [ -d "$sdk" ] && rm -rf "$sdk"
+    rm -f iPhoneOS8.0.sdk.tar.lzma
```
BrentDaMage, line 30: Why are we deleting the SDK just to redownload it?

platforms/macos/build.sh
```diff
@@ -5,40 +5,57 @@ set -e
...
-sdkver=1
-if ! [ -d "$x86_sdk" ] || ! [ -d "$arm64_sdk" ] || [ "$(cat sdkver 2>/dev/null)" != "$sdkver" ]; then
+sdkver=2
+if ! [ -d "$x86_64_sdk" ] || ! [ -d "$arm64_sdk" ] || ! [ -d "$old_sdk" ] || [ "$(cat sdkver 2>/dev/null)" != "$sdkver" ]; then
     printf '\nDownloading macOS SDKs...\n\n'
     (
     # for arm64
-    [ -d "$arm64_sdk" ] && rm -rf "$arm64_sdk"
+    [ -d "$arm64_sdk" ] && rm -rf "$arm64_sdk" &
+    rm -f MacOSX11.3.tar.bz2
```
BrentDaMage, line 30: Why are we removing the SDK just to redownload it?

Un1q32: If wget fails or the script is interrupted then there will be an incomplete or corrupted file.  Wget will save the file with a .1 at the end of the name and then tar will operate on the bad file.

BrentDaMage: So we're only redownloading the file if it's incomplete or corrupted?

Un1q32: Yea if theres a file there and sdkver isnt set correct then we assume its prolly a bad file.  I can't think of any other reason that could happen besides someone manually deleting sdkver, and in that case they prolly want the sdks redownloaded anyway.

BrentDaMage (approved):

PR #444, opened 2026-02-10
Armor and Farming

samdotjpg:
image

### Armor
- Beta style armor so every armor starts at "full effectiveness", and depending on the tier, it'll break faster.
- Better hurting and spill damage to the player
- Renders into GUI reading from right to left.
  - **DO NOT MIRROR THE ARMOR ICONS IN THE POCKET UI.** I **_know_** the original pocket UI reads from left to right (and beside the hearts), but mirroring it to accomplish that effect does not produce the same sprite. It also breaks texture pack compatibility by looking stupid. The compromise here has been to put it on the far right of the screen, saying as it's not necessary for that side to be used anyway due to no hunger bar.
### Farming
- Hoes, seeds, crops
## Miscellaneous Changes
- Second layer on humanoids
- Stairs drop stairs now. The old versions dropped their parent's item but that is stupid and nobody likes that.
- Changes Item::Tier to ToolItem::Tier because now we have ArmorItem::Tier. It's fucking confusing otherwise.
- Renamed a bunch of "field" variables to be what they are
  - Also removed a bunch of fields that will never be used and never were used in the original source code.
## Fixed Bugs
- Farmland not finding water sources and bonemeal responds to crops.
- Scaling issues on mirrored meshes

Thanks @Wilyicaro as per usual for being a frame of reference.

I got bored and wrote it like a charming little Minecraft update post do I get brownie points isn't it cute .  hello? did you see my messa

### DIFF
--- a/source/client/gui/Gui.cpp
+++ b/source/client/gui/Gui.cpp
@@ -455,7 +455,7 @@ void Gui::renderHearts(bool topLeft)
     }
 
 	int playerHealth = player->m_health;
-	int maxHealth = m_pMinecraft->m_pLocalPlayer->getMaxHealth();
+	int maxHealth = player->getMaxHealth();
 
 	for (int healthNo = 1; healthNo <= maxHealth; healthNo += 2)
 	{


samdotjpg:
I'll fix the MSVC errors when I'm home.

source/client/app/Minecraft.cpp
```diff
@@ -426,7 +426,7 @@ void Minecraft::handleBuildAction(const BuildActionIntention& action)
 				// @BUG: Hits sometimes pass through fire when done from above
 				//if (extinguished) break;
 
-				if (pTile != Tile::unbreakable || (player->field_B94 >= 100 && !m_hitResult.m_bUnk24))
```
BrentDaMage, line 429: I would like to keep things like this, since a lot of times the fields end up getting documented, and we are able to use this code as a frame of reference for doing so.

source/client/gui/Gui.cpp
```diff
@@ -193,7 +193,7 @@ void Gui::render(float f, bool bHaveScreen, int mouseX, int mouseY)
 		t.begin(0);
 		t.voidBeginAndEndCalls(true);
 
-		renderHearts(isTouchscreen);
+		renderHeartsAndArmor(isTouchscreen);
```
BrentDaMage: Can you put this in its own function? We have "renderHearts" "renderHunger" and "renderBubbles". I feel like combining one of them breaks that pattern.

source/client/model/models/HumanoidModel.hpp
```diff
@@ -22,11 +22,10 @@ class HumanoidModel : public Model
 	void setBrightness(float) override;
 
 public:
-	bool field_20;
-	ModelPart m_head, m_body, m_arm1, m_arm2, m_leg1, m_leg2;
+	Materials m_humanoidMaterials;
```
BrentDaMage: This should be protected or private.

BrentDaMage: Also, where the hell is `Materials` defined? For all of the other classes that follow this pattern of having their on materials, they would always have the `Materials` struct defined in the header.

source/client/renderer/TileRenderer.cpp
```diff
@@ -828,6 +886,23 @@ bool TileRenderer::tesselateCrossInWorld(Tile* tile, const TilePos& pos)
 	return true;
 }
 
+bool TileRenderer::tesselateRowInWorld(Tile* tile, const TilePos& pos)
+{
+	Tesselator& t = Tesselator::instance;
+
+	int color = getTileColor(tile, pos);
+
+	float r = float(GET_RED  (color)) / 255.0f;
```
BrentDaMage: I'm pretty sure we can initialize a Color from a 32-bit int. There's a constructor for it iirc.

source/world/tile/CropsTile.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+	if (diag || (hor && vert))
+	{
+		rate /= 2.0F;
+	}
+
+	return rate;
+}
+
+void CropsTile::growCropsToMax(Level* level, const TilePos& pos) {
```
BrentDaMage: Inconsistent formatting compared to the rest of the file.

source/world/tile/CropsTile.hpp
```diff
@@ -0,0 +1,24 @@
...
+
+public:
+	bool isSolidRender() const override;
+	bool isCubeShaped() const override;
+	eRenderShape getRenderShape() const override;
+	void tick(Level*, const TilePos& pos, Random*) override;
+	int getTexture(Facing::Name face, TileData data) const override;
+	void updateShape(const LevelSource* level, const TilePos& pos) override;
+	void spawnResources(Level*, const TilePos& pos, TileData data) override;
+
```
BrentDaMage, line 19: bonus new line

source/world/tile/FarmTile.cpp
```diff
@@ -94,23 +99,19 @@ void FarmTile::stepOn(Level* level, const TilePos& pos, Entity* pEnt)
 
 void FarmTile::tick(Level* level, const TilePos& pos, Random* random)
 {
-	int val = random->genrand_int32();
-
-	//@HUH: weird way of saying val % 5 == 0
-	if (val != 5 * (val / 5))
+	if (level->m_bIsClientSide || random->nextInt(5) != 0)
```
BrentDaMage: Make these separate if statements

GameMods.hpp
```diff
@@ -45,6 +45,7 @@
 //#define ENH_FACED_TERRAIN_PARTICLES 	   // Sets the TerrainParticle's texture depending on the face the block is being hit from. This is something Notch never did for whatever reason.
 #define ENH_NEW_LADDER_BEHAVIOR 	       // Use Java Beta 1.5 ladder behavior
 //#define ENH_GFX_COMPACT_UVS              // Use two unsigned 16-bit integers for UVs instead of a float Vec2
+//#define ENH_RENDER_HAND_ARMOR            // Render the right arm of the chestplate you are wearing on top of the GUI hand.
```
BrentDaMage: Call this the first-person hand instead of the GUI hand.

source/world/entity/Mob.cpp
```diff
@@ -881,12 +881,10 @@ void Mob::checkDespawn(Mob* nearestMob)
 		if (remWhenFar && distSqr > 9216.0f)
 			remove();
 
-		if (m_noActionTime <= 600)
-            m_noActionTime = 0;
-		else if (m_random.nextInt(800) == 0 && remWhenFar && distSqr >= 1024.0f)
```
samdotjpg, line 886: This is the reason entities never despawned by the way. This is effectively saying:
Has it been less than 600 seconds since the entity had action? Okay, reset action time, it's a closeby entity.
Has the entity been >= 1024 away from the player and a 1 in 800-ish chance occurs? Okay, remove.
But then, if neither of those are true, so if the entity hasn't been seen in less than 600 seconds, and the random number wasn't 1 in 800-ish, let's just reset the timer completely.
The actual java check is basically just "Is the entity not done any interactions in 600 ticks or more, 1024 away from the player or more dist sqr, and a diceroll of 1-in-800? Okay, remove it. Otherwise if it's close enough to the player, refresh its distance timer." Now monsters are despawning and not lingering forever.

BrentDaMage: Something interesting about this, I think it's possible actual PE had this bug, since during my testing with 0.3.X on my Android tablet, all of the zombies that would spawn at night would never seem to despawn.

source/client/gui/Gui.cpp
```diff
@@ -420,7 +421,37 @@ void Gui::renderMessages(bool bShowAll)
...
+
+			if (healthNo > armor)
+				blit(armorX, armorY, 16, 9, 9, 9, 0, 0);
+		}
+
+		armorX -= 8;
+	}
+}
+
+void Gui::renderHearts(bool topLeft)
```
BrentDaMage, line 424: renderHearts should come before renderArmor, just like in the header file

source/client/model/geom/Cube.cpp
```diff
@@ -22,6 +22,13 @@ Cube::Cube(ModelPart* a2, const TextureOffset& texOffs, float x, float y, float
 	y2 += g;
 	z2 += g;
 
+	field_29C = x1;
```
BrentDaMage: Okay, nah, you can remove these.

BrentDaMage (changes requested):

source/world/entity/Mob.cpp
samdotjpg: I do think it did have that bug. It feels very familiar, and the original Java code they adapted from had some unfortunate indentation that made them probably read the function incorrectly.

source/client/gui/Gui.cpp
```diff
@@ -454,8 +455,9 @@ void Gui::renderHearts(bool topLeft)
     }
 
 	int playerHealth = player->m_health;
+	int maxHealth = player->getMaxHealth();
 
-	for (int healthNo = 1; healthNo <= C_MAX_MOB_HEALTH; healthNo += 2)
```
samdotjpg, line 458: Ok, I think everything is done. Also, is C_MAX_MOB_HEALTH intentionally still available? It seems to be a hard cap over the actual getMaxHealth() to make sure no health goes over 15 unless heal is overwritten. idk

source/world/entity/Mob.cpp
BrentDaMage: Could we put a comment containing the old code? Possibly explaining what bug it caused, etc.

source/client/gui/Gui.cpp
BrentDaMage: Nuke C_MAX_MOB_HEALTH. It was probably before getMaxHealth existed in our code, but not 100% sure.

BrentDaMage (approved):

PR #446, opened 2026-02-11
World Creation Game Mode Switcher

samdotjpg:
- Adds a game mode switcher based on the positioning of Beta 1.8, the lowest JE version we can take from that has creative mode.
- Matches the button sizes to the rest of the UI, except on desktop.
- Cramped mode adjusted to fit, and keeps in mind the dynamic button sizes
- When we add more world options, we can replace the seed entry with it, or place it next to the game mode switcher.

Touchscreens:
Screenshot_20260211_094946
Desktop:
Screenshot_20260211_095156
Desktop (cramped):
Screenshot_20260211_095226
iPhone 3GS (cramped) ((thanks Uniq)):
IMG_0001

### DIFF
--- a/source/world/gamemode/GameType.hpp
+++ b/source/world/gamemode/GameType.hpp
@@ -16,41 +16,53 @@ enum GameType
 class GameTypeConv
 {
 public:
-	static const std::string& GameTypeToNonLocString(GameType type)
+static const std::string& GameTypeToNonLocString(GameType type)
+{
+	switch (type)
 	{
-		switch (type)
-		{
-		case GAME_TYPE_SURVIVAL:
-			static const std::string survival = "Survival";
-			return survival;
-		case GAME_TYPE_CREATIVE:
-			static const std::string creative = "Creative";
-			return creative;
-		case GAME_TYPE_ADVENTURE:
-			static const std::string adventure = "Adventure";
-			return adventure;
-		case GAME_TYPE_SPECTATOR:
-			static const std::string spectator = "Spectator";
-			return spectator;
-		default:
-			return Util::EMPTY_STRING;
-		}
+	case GAME_TYPE_SURVIVAL:
+	{
+		static const std::string survival = "Survival";
+		return survival;
 	}
+	case GAME_TYPE_CREATIVE:
+	{
+		static const std::string creative = "Creative";
+		return creative;
+	}
+	case GAME_TYPE_ADVENTURE:
+	{
+		static const std::string adventure = "Adventure";
+		return adventure;
+	}
+	case GAME_TYPE_SPECTATOR:
+	{
+		static const std::string spectator = "Spectator";
+		return spectator;
+	}
+	default:
+		return Util::EMPTY_STRING;
+	}
+}
 
-	static const std::string& GameTypeDescriptionToNonLocString(GameType type)
+static const std::string& GameTypeDescriptionToNonLocString(GameType type)
+{
+	switch (type)
 	{
-		switch (type)
-		{
-		case GAME_TYPE_SURVIVAL:
-			static const std::string survival = "Mobs, health and gather resources";
-			return survival;
-		case GAME_TYPE_CREATIVE:
-			static const std::string creative = "Unlimited resources, flying";
-			return creative;
-		case GAME_TYPE_ADVENTURE:
-		case GAME_TYPE_SPECTATOR:
-		default:
-			return Util::EMPTY_STRING;
-		}
+	case GAME_TYPE_SURVIVAL:
+	{
+		static const std::string survival = "Mobs, health and gather resources";
+		return survival;
 	}
+	case GAME_TYPE_CREATIVE:
+	{
+		static const std::string creative = "Unlimited resources, flying";
+		return creative;
+	}
+	case GAME_TYPE_ADVENTURE:
+	case GAME_TYPE_SPECTATOR:
+	default:
+		return Util::EMPTY_STRING;
+	}
+}
 };


source/client/gui/screens/CreateWorldScreen.cpp
```diff
@@ -117,26 +127,46 @@ void CreateWorldScreen::init()
...
+
+	m_gameMode = GAME_TYPE_SURVIVAL;
 }
 
 void CreateWorldScreen::render(float f)
 {
 	renderBackground();
 	Screen::render(f);
 
+	const std::string gameTypeStr = GameTypeConv::GameTypeDescriptionToNonLocString(m_gameMode);
```
samdotjpg, line 154: I'm apprehensive about this. Right now it's grabbing a string every single render frame, but caching it requires you to, well, cache a string. Advisal of the council please

source/client/gui/screens/CreateWorldScreen.cpp
```diff
@@ -117,26 +127,46 @@ void CreateWorldScreen::init()
...
+		m_btnGameMode.m_yPos = m_textSeed.m_yPos + 30;
+		m_btnCreate.m_width -= 50;
+		m_btnBack.m_width -= 50;
 	}
+
+	// calculate after crush
+	m_btnGameMode.m_xPos = m_width / 2 - m_btnGameMode.m_width / 2;
+	m_btnCreate.m_xPos = m_width / 2 - m_btnCreate.m_width - 5;
+
+	m_gameMode = GAME_TYPE_SURVIVAL;
```
samdotjpg, line 146: Default game mode as survival?

source/client/gui/screens/CreateWorldScreen.cpp
```diff
@@ -117,26 +127,46 @@ void CreateWorldScreen::init()
...
 	drawString(*m_pFont, "Seed for the World Generator",  m_textSeed.m_xPos, m_textSeed.m_yPos - 10, 0xDDDDDD);
-	drawString(*m_pFont, "Leave blank for a random seed", m_textSeed.m_xPos, m_textSeed.m_yPos + 22, 0x999999);
+	
+	bool crampedMode = CRAMPED();
+
+	// Extra text not necessarily needed
+	if (!crampedMode)
+	{
+		drawString(*m_pFont, "Leave blank for a random seed", m_textSeed.m_xPos, m_textSeed.m_yPos + 22, 0x999999);
+		int textCenteredX = m_btnGameMode.m_xPos + (m_btnGameMode.m_width / 2) - (m_pFont->width(gameTypeStr.c_str()) / 2);
```
samdotjpg: it was bogging up the width of the file to have this all in one large check but i can make it a large check if need be

source/client/gui/screens/CreateWorldScreen.cpp
bluepilledgreat: Unless it causes major performance issues, it should be fine for now.

source/client/gui/screens/CreateWorldScreen.cpp
```diff
@@ -117,26 +127,46 @@ void CreateWorldScreen::init()
...
 
+	const std::string gameTypeStr = GameTypeConv::GameTypeDescriptionToNonLocString(m_gameMode);
+	m_btnGameMode.setMessage("Game Mode: " + GameTypeConv::GameTypeToNonLocString(m_gameMode));
+
 	drawCenteredString(*m_pFont, "Create New World", m_width / 2, CRAMPED() ? 10 : 30, 0xFFFFFF);
 	drawString(*m_pFont, "World name",                    m_textName.m_xPos, m_textName.m_yPos - 10, 0xDDDDDD);
 	drawString(*m_pFont, "Seed for the World Generator",  m_textSeed.m_xPos, m_textSeed.m_yPos - 10, 0xDDDDDD);
-	drawString(*m_pFont, "Leave blank for a random seed", m_textSeed.m_xPos, m_textSeed.m_yPos + 22, 0x999999);
+	
+	bool crampedMode = CRAMPED();
```
bluepilledgreat, line 161: Why is this a macro? It would be better as a static function.

samdotjpg: Didn't make that. Just using what was there.

source/client/gui/screens/CreateWorldScreen.cpp
Un1q32: yes

source/client/gui/screens/CreateWorldScreen.cpp
BrentDaMage: Honestly, this is my fault for not improving the decompiled Mojang code. What we should be doing, is initializing those strings in the function as static members of the `GameTypeConv` class, then just have those functions return const references to them. Then obviously at that point, gameTypeStr should become a const reference as well.

source/client/gui/screens/CreateWorldScreen.cpp
BrentDaMage: yeah

source/client/gui/screens/CreateWorldScreen.cpp
BrentDaMage: you can also just have it do
```cpp
if (crampedMode)
    return;
```

samdotjpg: Good idea

source/world/gamemode/GameType.hpp
```diff
@@ -15,35 +16,42 @@ enum GameType
 class GameTypeConv
 {
 public:
-	static std::string GameTypeToNonLocString(GameType type)
+	static const std::string& GameTypeToNonLocString(GameType type)
 	{
+		static std::string survival = "Survival";
```
BrentDaMage: okay, I mean his technically works I guess. Can you make them const as well, and can you have case and return be on the same line? thanks

source/world/gamemode/GameType.hpp
```diff
@@ -15,35 +16,42 @@ enum GameType
...
 		default:
-			return "Undefined";
+			return Util::EMPTY_STRING;
 		}
 	}
 
-	static std::string GameTypeDescriptionToNonLocString(GameType type)
+	static const std::string& GameTypeDescriptionToNonLocString(GameType type)
 	{
+		static std::string survival = "Mobs, health and gather resources";
```
BrentDaMage: do the same down here

samdotjpg:
done i think

BrentDaMage (approved):

BrentDaMage (approved):

PR #472, opened 2026-02-15
Network Protocol V6

BrentDaMage:
* Added PlayerActionPacket
* Only sent when using items


Un1q32:
Don't forget to update the version again.

BrentDaMage:
> Don't forget to update the version again.

Thank you.

PR #475, opened 2026-02-15
Fix mismatched allocs

Un1q32:
ImageData.m_data must be allocated with malloc/free because it is in stb_image and also in the emscripten loadImage implementation.

### DIFF
--- a/source/client/renderer/texture/TextureAtlas.cpp
+++ b/source/client/renderer/texture/TextureAtlas.cpp
@@ -73,7 +73,7 @@ bool TextureAtlas::build()
 
 void TextureAtlas::_init()
 {
-    free(m_texture.m_imageData.m_data);
+    m_texture.m_imageData.release();
 
     int size = getWidth() * getHeight() * 4;
     uint8_t *mem = (uint8_t *)malloc(size);


BrentDaMage:
Why?

Un1q32:
You can't free() something allocated with new, and you can't delete something allocated with malloc.  stb_image and emscripten allocate data with malloc, so it has to be free'd, which it already was in most areas of the code.  Since some parts of the code always free the data with free, other parts that allocate the data have to use malloc as well so those parts that free with free still work.

Un1q32:
I discovered this issue by running asan.  Exiting the game throws a mismatched alloc error because data allocated with new[] is freed with free().  Changing the free() to a delete[] makes it error earlier as it delete[]'s data allocated with malloc by stb_image.

source/client/renderer/texture/TextureAtlas.cpp
```diff
@@ -73,10 +73,13 @@ bool TextureAtlas::build()
 
 void TextureAtlas::_init()
 {
-    SAFE_DELETE_ARRAY(m_texture.m_imageData.m_data);
+    free(m_texture.m_imageData.m_data);
```
BrentDaMage: `m_texture.m_imageData.release()` should be called here instead.

BrentDaMage: You also probably forgot to use `free` in `ImageData::release()`.

BrentDaMage (changes requested):

source/client/renderer/texture/TextureAtlas.cpp
Un1q32: ImageData::release already used free().

PR #485, opened 2026-02-17
Chat & Pause buttons for touch screens

Vimd0:
Screenshot_20260216_225844_ReMinecraftPE
Screenshot_20260216_225855_ReMinecraftPE

- Chat and Pause button
- Pocket theme now exclusively have 6 slots
- Updated guitouch to have reversed gradient for the Play, Options, and Multiplayer buttons

### DIFF
--- a/source/client/gui/Gui.cpp
+++ b/source/client/gui/Gui.cpp
@@ -308,6 +308,7 @@ void Gui::handleClick(int clickID, int mouseX, int mouseY)
 	if (clickID != 1)
 		return;
 
+	// @TODO: add InGamePlayScreen at some point
 	if (m_pMinecraft->isTouchscreen())
     {
 		int cenX = GuiWidth / 2;


source/client/gui/screens/IngameBlockSelectionScreen.cpp
```diff
@@ -187,31 +194,21 @@ bool IngameBlockSelectionScreen::isInsideSelectionArea(int x, int y)
...
 	m_btnCraft.m_xPos = 0;
 	m_btnCraft.m_yPos = 0;
-	_addElement(m_btnCraft);*/
+	/*if (m_pMinecraft->isTouchscreen())
+	{
+		_addElement(m_btnCraft);
+	}*/
 
-	/*m_btnArmor.m_width = 40;
+	m_btnArmor.m_width = 40;
```
BrentDaMage: These do nothing, why uncomment them?

source/client/gui/screens/IngameBlockSelectionScreen.cpp
```diff
@@ -187,31 +194,21 @@ bool IngameBlockSelectionScreen::isInsideSelectionArea(int x, int y)
...
-	m_btnChat.m_width = 25;
-	m_btnChat.m_xPos = 0;
-	m_btnChat.m_yPos = 0;
-	if (m_pMinecraft->isTouchscreen())
-	{
-		_addElement(m_btnChat);
-	}
-	
-	/*m_btnCraft.m_width = 40;
+	m_btnCraft.m_width = 40;
```
BrentDaMage: These do nothing, why uncomment them?

source/client/gui/screens/IngameBlockSelectionScreen.cpp
```diff
@@ -109,7 +106,17 @@ IngameBlockSelectionScreen::IngameBlockSelectionScreen() :
 	addCreativeItem(Tile::netherrack->m_ID);
 	addCreativeItem(Tile::soulSand->m_ID);
 	addCreativeItem(Tile::glowstone->m_ID);
+
 	addCreativeItem(Tile::web->m_ID);
+	addCreativeItem(Tile::ice->m_ID);
```
BrentDaMage: I don't think we have the space for this without it going way off-screen on iPhones, but whatever.

source/client/gui/Gui.cpp
```diff
@@ -307,6 +308,28 @@ void Gui::handleClick(int clickID, int mouseX, int mouseY)
 	if (clickID != 1)
 		return;
 
+	if (m_pMinecraft->isTouchscreen())
```
BrentDaMage, line 312: Is there really no way to put an actual button here? Why this?

source/client/gui/screens/IngameBlockSelectionScreen.cpp
Vimd0: fixed

source/client/gui/screens/IngameBlockSelectionScreen.cpp
Vimd0: fixed

source/client/gui/screens/IngameBlockSelectionScreen.cpp
Vimd0: I accidentally kept these items I use for personal builds

source/client/gui/Gui.cpp
Vimd0: We need this because it's the only way to handle clicks for elements in game

BrentDaMage: Afaik, PE handles this using `InGamePlayScreen`, which renders the buttons as `GuiElements`.

Vimd0: Would you like me to do this another way? I feel like this is the best way to do this

BrentDaMage: This is the easiest way to do this, for sure. I think it's probably fine how it is now, but we should put a TODO message by it noting that we should add InGamePlayScreen at some point.

Vimd0: For sure

PR #486, opened 2026-02-17
steal grasscolor.png

Un1q32:
i doubt this is copyrightable at all, and even if it is i doubt this will be the one deciding factor that makes mojang kill us all.


BrentDaMage:
I would advise you not name the PR this, but whatever.

PR #489, opened 2026-02-18
Fix default UI theme, also fix Minecraft::isTouchscreen always being false

Un1q32:


source/client/app/Minecraft.cpp
```diff
@@ -915,6 +915,7 @@ void Minecraft::update()
 void Minecraft::init()
 {
 	m_pRakNetInstance = new RakNetInstance;
+	m_bIsTouchscreen = platform()->isTouchscreen();
```
BrentDaMage: This is already done in `NinecraftApp::_initInput()`

BrentDaMage: You should be good to just remove this.

PR #509, opened 2026-02-22
fix use-after-free

Un1q32:
According to clang this causes a UAF because the std::string is destroyed, fixed by declaring the temporary std::string in a different scope.

### DIFF
--- a/platforms/sdl/sdl1/main.cpp
+++ b/platforms/sdl/sdl1/main.cpp
@@ -86,36 +86,26 @@ static void initGraphics()
 
 static std::string getStoragePath()
 {
-    // Doing this as a c-string because worst-case, an SDK
-    // will return a nullptr instead of an empty string
-    const char* pathBase;
+    std::string path;
+    const char *tmp;
 #ifdef _WIN32
-    pathBase = getenv("APPDATA");
+    tmp = getenv("APPDATA");
+    if (tmp)
+        path = tmp;
 #else
-    std::string tmp;
-    const char *xdg_data = getenv("XDG_DATA_HOME");
-    if (xdg_data)
-        pathBase = xdg_data;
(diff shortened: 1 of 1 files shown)


platforms/sdl/sdl1/main.cpp
```diff
@@ -104,7 +105,10 @@ static std::string getStoragePath()
             pathBase = ""; // current working directory
         }
         else
-            pathBase = ((std::string)xdg_data + "/.local/share").c_str();
```
BrentDaMage, line 107: in what way is this a use-after-free?

Un1q32: according to clang, the std::string is backing the char*, and when the std::string is destroyed (immediately in this case) the char* is deallocated.

BrentDaMage: right, but the assignment operator should be copying the contents of the .c_str() result to pathBase before xdg_data is freed, no?

Un1q32: Copying to what memory?  pathBase is just a pointer it doesn't have any memory.  std::string creates a memory buffer and assigns the pointer to that buffer to pathBase, but when the std::string is destroyed that buffer is deallocated.

BrentDaMage: image
the xdg_data.c_str() pointer is used to copy the data from xdg_data to pathBase

BrentDaMage: Never mind, just realized that pathBase is not an std::string. My bad.

Un1q32: So do you understand now? can i mark this convo as resolved?

Un1q32:
done

PR #512, opened 2026-02-23
macOS Fixes

BrentDaMage:
* Fixed building on Power Mac G5
* Fixed external storage path on macOS to go to `~/Library/Application Support/.reminecraftpe` instead of `~/.local/.share`


platforms/sdl/sdl1/main.cpp
```diff
@@ -92,6 +92,10 @@ static std::string getStoragePath()
     tmp = getenv("APPDATA");
     if (tmp)
         path = tmp;
+#elif defined(MC_PLATFORM_MAC)
```
Un1q32, line 95: why not just use `__APPLE__` here?

BrentDaMage: because `__APPLE__` is likely defined on iOS, and because `MC_PLATFORM_MAC` is clearer

Un1q32: ig maybe if someone wants to make an sdl2 iOS port

PR #515, opened 2026-02-24
Input Hotswap & Misc

### LINKED ISSUE #413: Input hotswapping/multi-support
For example, right now if you are using a touchscreen device such as an android that also has a built-in controller, there is no way to do most actions with the controller. With a controller, you can break, place, swap inventory positions, open inventory, do inventory management, open chat, open and adjust settings, and change perspective. You can't move, jump, or look.

### LINKED ISSUE #595: Let controller support be togglable
I often have my controller plugged into my computer, and when I try to play nbcraft, it defaults to the controller over keyboard and mouse.
There's a toggle in the controls menu and everything, it's just disabled.

Image

The toggle is probably just for aesthetic purposes, but would be really cool if it had a use (or atleast a command-line argument to force a input type or something like that).

Wilyicaro:
- Added input hotswap, tested between controller and kbm for now
- Added a HashMap implementation in common, and implemented in some places
- Now, Console UI Theme tooltips will be accurate
- Now, sneaking with the controller will be toggleable
- Now, the UI Theme will be applied when exiting the options screen
- Now, the font rendering will be faster
- Now, Vertical Layout will always navigate to the next element when scrolling down
- Fixed Skeleton arrow shooting not accounting the head height
- Fixed item swapping not calling slot changed, not changing the crafting result
- Fixed Vertical Layout labels being selectable
- Fixed not being able to navigate if there aren't selected elements

## TODO
- [ ] Universal Keyboard codes
- [ ] Allow selecting different controllers (not sure about that)
- [x] Finish some Console UIs


platforms/input/xinput/GameControllerHandler_xinput.cpp
```diff
@@ -95,9 +89,9 @@ void GameControllerHandler_xinput::refresh()
     {
         XINPUT_STATE& inputState = m_inputStates.m_inputState[id];
         bool joinGameAlreadyFired = false;
-        for (ButtonIDMap::const_iterator it = m_buttonIdMap.begin(); it != m_buttonIdMap.end(); it++)
+        for (ButtonIDMap::Iterator it = m_buttonIdMap.begin(); it != m_buttonIdMap.end(); it++)
```
BrentDaMage: I'd really prefer that our HashMap class have the same general interface semantics as `std::unordered_map`

BrentDaMage: So this means `Iterator` being `iterator` `it.key()` being `it->first`, etc. This will make it way easier to quickly change `std::map`s to hash maps.

platforms/sdl/base/AppPlatform_sdl.cpp
```diff
@@ -492,3 +492,54 @@ Keyboard::KeyState AppPlatform_sdl::GetKeyState(uint8_t state)
 		return Keyboard::DOWN;
 	}
 }
+
+GameController::EngineButtonID AppPlatform_sdl::GetEngineButton(uint8_t button)
+{
+	switch (button)
+	{
+	case SDL_CONTROLLER_BUTTON_A:
```
BrentDaMage: Could you make the case return take up one line?

platforms/sdl/base/AppPlatform_sdl.cpp
```diff
@@ -492,3 +492,54 @@ Keyboard::KeyState AppPlatform_sdl::GetKeyState(uint8_t state)
 		return Keyboard::DOWN;
 	}
 }
+
+GameController::EngineButtonID AppPlatform_sdl::GetEngineButton(uint8_t button)
```
BrentDaMage: This could just be `_getEngineButton`, and only exist in the CPP file. No need to declare it in the header or as static and whatnot.

source/client/app/Minecraft.cpp
```diff
@@ -361,17 +372,17 @@ bool Minecraft::isOnlineClient() const
 
 bool Minecraft::isTouchscreen() const
 {
-	return m_bIsTouchscreen;
+	return IInputHolder::activeType == IInputHolder::TOUCHSCREEN;
```
BrentDaMage: This should **really** be a static getter somewhere *worst-case*.

BrentDaMage: Also, I'm not a fan of having the enum inside of `IInputHolder`, since we've never put enums in interfaces before. Additionally, this makes it difficult to tell what exactly the enum is representing. Is it an input holder type?

source/client/app/Minecraft.cpp
```diff
@@ -361,17 +372,17 @@ bool Minecraft::isOnlineClient() const
...
 bool Minecraft::useSplitControls() const
 {
-	return !m_bIsTouchscreen || getOptions()->m_splitControls.get();
+	return !isTouchscreen() || getOptions()->m_splitControls.get();
 }
 
 bool Minecraft::useController() const
 {
-	return m_pPlatform->hasGamepad() && getOptions()->m_bUseController.get();
+	return m_pPlatform->hasGamepad() && (IInputHolder::activeType == IInputHolder::CONTROLLER || getOptions()->m_bUseController.get());
```
BrentDaMage: Do we even need to check `AppPlatform::hasGamepad()` or could we just assume the input holder type will always be correct?

source/client/app/Minecraft.cpp
```diff
@@ -550,93 +564,132 @@ void Minecraft::tickInput()
...
+		//Clearing, so the events don't accumulate
+		Keyboard::reset();
 
-		m_pLocalPlayer->m_pMoveInput->setKey(keyCode, bPressed);
-
-		if (bPressed)
+		while (GameControllerManager::next())
 		{
-			m_pGui->handleKeyPressed(keyCode);
+			if (!useController()) continue;
```
BrentDaMage: Maybe just don't bother iterating at all if `useController` returns false. If the return value can change mid-loop, you should probably just call `break` instead of `continue`.

source/client/app/Minecraft.cpp
```diff
@@ -550,93 +564,132 @@ void Minecraft::tickInput()
...
-			}
-#endif
 		}
+	}
 
-		if (getOptions()->field_19)
-			continue;
+	for (int i = 0; i < m_pGui->getNumUsableSlots(); i++)
+	{
+		while (getOptions()->getControlMapping(eControlMappingIndex(KM_SLOT_1 + i)).consume())
```
BrentDaMage: Maybe call `getControlMapping` `getInputMapping`?

source/client/app/Minecraft.cpp
```diff
@@ -550,93 +564,132 @@ void Minecraft::tickInput()
...
-		// @TODO: Replace with KeyboardBuildInput
-		if (!useController() && getTimeMs() - field_2B4 <= 200)
+	while (getOptions()->getControlMapping(KM_MENU_PAUSE).consume())
+	{
+		handleBack(false);
+	}
+
+	while (getOptions()->getControlMapping(KM_DROP).consume())
+	{
+		ItemStack& item = m_pLocalPlayer->m_pInventory->getSelected();
```
BrentDaMage: I'm not a fan of this logic existing in `Minecraft`. I feel like it should go in `LocalPlayer` or something.

source/client/app/Minecraft.cpp
```diff
@@ -847,6 +900,9 @@ void Minecraft::tick()
 			m_pScreen->tick();
 
 		Multitouch::reset();
+
+		// Actually pause the game, because fuck bedrock edition
+		m_bIsGamePaused = !isOnline() || (m_pLevel && m_pLevel->m_players.size() == 1 && m_pScreen && m_pScreen->isPauseScreen());
```
BrentDaMage: Wouldn't this just freeze the game if the player doesn't have a RakNet instance?

source/client/app/Minecraft.cpp
BrentDaMage: Could we remove the hack at this point, or would this still break it?

source/client/gui/components/TextBox.cpp
```diff
@@ -205,26 +205,26 @@ char TextBox::guessCharFromKey(int key) {
 
 #endif
 
-void TextBox::handleButtonPress(Minecraft* pMinecraft, int key)
+void TextBox::handleControlPress(Minecraft* pMinecraft, const ControlBind& bind)
```
BrentDaMage: I'm not a fan of `control` here. Maybe we could workshop a name?

source/client/gui/screens/OptionsScreen.cpp
```diff
@@ -164,6 +164,16 @@ void OptionsScreen::handleScrollWheel(float force)
 	m_pList->handleScrollWheel(force);
 }
 
+bool OptionsScreen::isInvalid(Minecraft* mc)
```
BrentDaMage: Smart

source/client/gui/screens/OptionsScreen.cpp
```diff
@@ -164,6 +164,16 @@ void OptionsScreen::handleScrollWheel(float force)
...
+{
+	if (mc->getOptions()->getUiTheme() == UI_CONSOLE)
```
BrentDaMage: I'd prefer if the actual changing of the current screen were handled in a `tick` function within the current Screen. Then `isInvalid` could just be `const`. Additionally, the name `isInvalid` implies that the function is just returning the result of a bool, not modifying the current screen, etc.

source/client/gui/screens/OptionsScreen_Console.cpp
```diff
@@ -134,6 +144,7 @@ OptionHeader_Console::OptionHeader_Console(const std::string& text)
 	: m_text(text)
 {
 	m_height = 22;
+	setNavigable(false);
```
BrentDaMage, line 157: My bad

source/client/gui/screens/OptionsScreen.cpp
BrentDaMage: Perhaps the `tick` function checks if the current screen should be invalidated. If so, it pushes a new screen, then calls a `Screen::invalidateScreen` or `Screen::invalidate` function. This function could change a private member variable which `isInvalid` could return.

source/client/app/Minecraft.cpp
BrentDaMage: 0.12.1 has `MinecraftInputHandler::getButtonId(const std::string&)` and passes something like `button.menu_up`. They likely have a `ButtonId` typedef or something, which ends up getting compiled as an `int16_t`. This would be accessed like so `getInput()->getButtonId("button.menu_up")`.

source/client/gui/Gui.hpp
```diff
@@ -65,7 +65,7 @@ class Gui : public GuiComponent
 	bool isInside(int mx, int my);
 	void handleClick(int id, int mx, int my);
 	void handleScrollWheel(bool down);
-	void handleKeyPressed(int keyCode);
+	void handleControlPressed(const ControlBind&);
```
BrentDaMage: Would prefer if this were called `handleButtonPress`, since button can apply to mice, keyboards, controllers, etc. Additionally, we have code that checks if shift or control are pressed, and this may be a tad confusing.

source/client/gui/Screen.cpp
```diff
@@ -139,55 +139,55 @@ void Screen::init(Minecraft* pMinecraft, int width, int height)
 	m_pFont = pMinecraft->m_pFont;
 
 	// Apply UI theme to current Screen based on user preference
-	UITheme userTheme = m_pMinecraft->getOptions()->getUiTheme();
+	UITheme userTheme = pMinecraft->getOptions()->getUiTheme();
 	// We don't bother applying the console theme automatically for generic screens because this completely fucks scaling
 	if (m_uiTheme == UI_UNIVERSAL || (m_uiTheme == UI_GENERIC && userTheme != UI_CONSOLE))
-		m_uiTheme = m_pMinecraft->getOptions()->getUiTheme();
+		m_uiTheme = userTheme;
```
BrentDaMage, line 146: Thanks, I was meaning to fix this, but I guess I forgot to.

source/client/player/input/GameControllerManager.cpp
```diff
@@ -63,6 +116,9 @@ void GameControllerManager::feedStickX(GameController::StickID stickId, bool tou
 	stickValuesX[index] = x; // LCE has a deadzone of 20000, ~0.3f on 360
 
 	inReset = false;
+
+	if (x)
```
BrentDaMage: Could you manually write this out as `x != 0.0f`? Same below.

source/client/player/input/KeyboardInput.cpp
```diff
@@ -29,19 +29,33 @@ void KeyboardInput::releaseAllKeys()
...
-	else if (m_pOptions->getKey(KM_LEFT)     == eventKey) index = INPUT_LEFT;
-	else if (m_pOptions->getKey(KM_RIGHT)    == eventKey) index = INPUT_RIGHT;
-	else if (m_pOptions->getKey(KM_JUMP)     == eventKey) index = INPUT_JUMP;
-	else if (m_pOptions->getKey(KM_SNEAK)    == eventKey) index = INPUT_SNEAK;
-
-	if (index == -1)
+	int index;
+	switch (ctrl)
+	{
+	case KM_FORWARD:
```
BrentDaMage: I'd recommend cramming these onto one line each.

source/client/renderer/Font.hpp
```diff
@@ -49,6 +50,7 @@ class Font
 	int field_0; 
 	int m_charWidthInt[256];
```
BrentDaMage: Could u make `256` a macro?

source/common/utility/HashMap.hpp
```diff
@@ -0,0 +1,413 @@
...
+    {
+        size_t hash = 5381;
+        while (*key)
+            hash = ((hash << 5) + hash) + size_t(*key++);
+        return hash;
+    }
+};
+
+template<typename TKey, typename TValue, typename THash = HashFunction<TKey> >
+class HashMap
```
BrentDaMage, line 98: Is this your own impl, or did this come from somewhere?

BrentDaMage (changes requested):

source/client/app/Minecraft.cpp
Wilyicaro: That's temporary, my plan is making this non-static for the splitscreen (member of `Minecraft`), but this will need other changes, considering `Mouse`, `Keyboard` and `GameControllerManager` are static.
But yeah, it's an input holder type for now, I wouldn't change this while splitscreen isn't a thing, do u really want to make changes?

source/client/app/Minecraft.cpp
Wilyicaro: You're right, the input holder type should be always correct, but I think that I would need to use a controller disconnect event to change the input back to default (kbm or touchscreen), if not, it would be stuck at the controller input type while you don't use any input
It's worth mentioning `Options::m_bUseController` is being used for locking the controller input, not sure if that's clear at all

source/common/utility/HashMap.hpp
Wilyicaro: It's my own impl

source/client/renderer/Font.hpp
Wilyicaro: Alr

source/client/player/input/KeyboardInput.cpp
Wilyicaro: Sure

platforms/input/xinput/GameControllerHandler_xinput.cpp
Wilyicaro: Alr

source/client/app/Minecraft.cpp
Wilyicaro: This is remaining from old code, I'll remove it

source/client/app/Minecraft.cpp
Wilyicaro: I don't think so, considering it's better to call this after the control mappings are updated in `Minecraft::tickInput`, and this is handled in `Minecraft` even in newer Java versions

source/client/gui/screens/OptionsScreen.cpp
Wilyicaro: I think renaming this to `Screen::validate`, inverting its return value and calling this also in the ticking should be enough
This can be used for the creative menu in the future

source/client/gui/Gui.hpp
Wilyicaro: So, `ControlBind` will be called `ButtonBind` instead?

Cjnator38:
> [ ] Finish some Console UIs

Game mode selector included, right?

Wilyicaro:
> > [ ] Finish some Console UIs
> 
> Game mode selector included, right?

yea

source/client/gui/screens/inventory/ContainerScreen.cpp
```diff
@@ -13,7 +13,7 @@ ContainerScreen::ContainerScreen(ContainerMenu* menu) :
     m_topPos(0),
     m_timeSlotDragged(0)
 {
-    m_uiTheme = UI_UNIVERSAL;
+    m_themeSelection = UI_UNIVERSAL;
```
BrentDaMage: I'm not a fan of this variable name.

BrentDaMage: What's the difference between `m_themeSelection` and `m_uiTheme`? Is one the user preference, and the other the actual theme of the Screen? If so, how about `m_screenTheme` and `m_userTheme`?

source/client/gui/Screen.cpp
```diff
@@ -47,7 +47,9 @@ Screen::Screen()
 	m_bRenderPointer = false;
 	m_lastTimeMoved = 0;
 	m_cursorTick = 0;
-	m_uiTheme = UI_GENERIC;
+	m_themeSelection = UI_GENERIC;
+	m_uiTheme = UI_POCKET;
+	m_bUniversalUiTheme = false;
```
BrentDaMage: Why does this bool need to be here?

source/client/renderer/Font.hpp
```diff
@@ -13,6 +13,8 @@
 #include "renderer/MaterialPtr.hpp"
 #include "client/renderer/renderer/Tesselator.hpp"
 
+#define FONT_CHARS_AMOUNT (256)
```
BrentDaMage: Can you put `C_` in front of this?

source/world/item/ItemStack.cpp
```diff
@@ -158,7 +158,7 @@ CompoundTag* ItemStack::getNetworkUserData() const
 {
 	CompoundTag* userData = new CompoundTag();
 	CompoundTag::NamedTagMap& tags = m_userData->rawView();
-	for (CompoundTag::NamedTagMap::Iterator it = tags.begin(); it != tags.end(); it++)
+	for (CompoundTag::NamedTagMap::iterator it = tags.begin(); it != tags.end(); it++)
```
BrentDaMage: This should be a `const_iterator`, no? I think it was originally as well.

BrentDaMage (changes requested):

source/client/app/Minecraft.cpp
BrentDaMage: Assuming you're going to make this non-static in this PR, great. Otherwise, I'd like it to be in a static function that isn't in the interface.

source/client/app/Minecraft.cpp
BrentDaMage: Yeah, I could tell that the "Use Controller" option was repurposed into disabling controller input. I'd recommend hiding this option entirely, or disabling it on console platforms. I think you can use `MC_PLATFORM_CONSOLE`.

source/client/gui/screens/inventory/ContainerScreen.cpp
BrentDaMage: Or perhaps `m_screenTheme` and `m_globalTheme`?

Wilyicaro: `m_themeSelection` uses another enum for updating the screen UI Theme, cuz before `UI_GENERIC` and `UI_UNIVERSAL` were updating once the UI Theme, causing issues when changing themes

source/client/gui/Screen.cpp
Wilyicaro: I explained that above

source/client/gui/screens/inventory/ContainerScreen.cpp
BrentDaMage: Is there a better variable name we could use that could help distinguish this between the UI theme? Perhaps we call it "m_screenType" and have `SCREEN_UNIVERSAL`, etc.? idk, just a thought

source/client/gui/Screen.cpp
BrentDaMage: Couldn't we just check the value of `m_themeSelection` though?

Wilyicaro: bruh, I forgot removing that

source/client/app/Minecraft.cpp
BrentDaMage: Alright then.

source/client/gui/Gui.hpp
BrentDaMage: Perhaps call this `ButtonEventData`? We're getting rid of the class at some point anyways, right?

platforms/input/xinput/GameControllerHandler_xinput.cpp
BrentDaMage: It looks like we're still doing `.key()` and `.value()` instead of `->first` and `->second`.

source/client/gui/screens/inventory/ContainerScreen.cpp
Wilyicaro: I'll think in a better name, that was just the first name that came to my mind, so I could fix the mentioned issue

source/world/item/ItemStack.cpp
Wilyicaro: Not really, as `userData` isn't const, nor are `tags`

source/client/gui/Gui.hpp
Wilyicaro: `ControlBind` is also used in `ControlMapping`, so that name wouldn't fit

source/client/app/Minecraft.cpp
Wilyicaro: In fact, it doesn't disable the controller input, it's kinda the opposite, it forces to use it

platforms/input/xinput/GameControllerHandler_xinput.cpp
Wilyicaro: I wouldn't change here, considering this will be useful when changing from `std::map`s to hash maps, changing here would be useless, as that change was already done.

Wilyicaro: But if you really want to replace every `.key()` and `.value()` usage, let me know

source/world/item/ItemStack.cpp
BrentDaMage: The tags are const, and while userData isn't const, the iterator value is copied.

source/client/gui/Gui.hpp
BrentDaMage: Then please call them InputMappings instead or something. Maybe call `ControlBind` `InputEvent`? idk

source/client/app/Minecraft.cpp
BrentDaMage: Oh wait, why?

source/client/app/Minecraft.cpp
Wilyicaro: Why not `getButtonMapping` ?

platforms/input/xinput/GameControllerHandler_xinput.cpp
BrentDaMage: I would prefer, that, yeah. That way it works like `std::unordered_map` and other devs won't get confused when interfacing with it or reading pre-existing code. I'd prefer we just have one way to access it.

source/client/app/Minecraft.cpp
BrentDaMage: Sure

source/client/app/Minecraft.cpp
Wilyicaro: > Oh wait, why?

It might be useful, mainly for splitscreen

BrentDaMage: When we add split-screen, wouldn't we just have a way to assign an input method to each player instead? I guess this works in the mean time, but its use isn't super apparent to an average user.

source/client/gui/Gui.hpp
Wilyicaro: It would be weird `ButtonMapping`/`InputMapping` having a member called event

Wilyicaro: For me, `ButtonBind` is fine

BrentDaMage: What the hell even is a ControlBind? It just contains an ID for a keystroke, and an ID for a GameController button? The name is already unclear, and the contents are already odd.

Wilyicaro: > What the hell even is a ControlBind? It just contains an ID for a keystroke, and an ID for a GameController button? The name is already unclear, and the contents are already odd.

What more did you expect? It's a bind, I couldn't expect more

Wilyicaro: The name would only be clearer if it were `KeycodeAndGameControllerButtonId` lmao

BrentDaMage: Could we not just have two maps, one for keyboard button binds, and one for controller button binds?

Wilyicaro: > Could we not just have two maps, one for keyboard button binds, and one for controller button binds?

I don't think that idea would be efficient, considering that `ButtonMapping` would still have to exist, but without the bind, if you're talking about creating two other maps just to store the keyboard and controller binds. Now, if you're thinking of having a `KeyMapping` and a `ControllerButtonMapping`, I think it's even less versatile...

Wilyicaro: Also, what's wrong with the current struct? I can't think of a reason.

BrentDaMage: Can we just call this `ButtonPressEvent` or something? Or `ButtonInputEvent`? We're calling this a binding but it doesn't seem to actually be a binding, it just contains a keyboard button ID and a controller button ID.

BrentDaMage: > Also, what's wrong with the current struct? I can't think of a reason.

I understand now why we have the struct, but I think the name does a bad job of conveying its purpose.

Wilyicaro: > Can we just call this `ButtonPressEvent` or something? Or `ButtonInputEvent`? We're calling this a binding but it doesn't seem to actually be a binding, it just contains a keyboard button ID and a controller button ID.

that's less clear, why a button press/input event would've just a keyboard button ID and a controller button ID?
Also, why a mapping would store an event?

Perhaps, `ButtonInfo`?

BrentDaMage: In what way is this a mapping? Are we mapping keyboard button ids to game controller IDs or something? I guess I don't understand the point of this struct then. I thought it was to handle simultaneous input from a keyboard and game controller.

Wilyicaro: > In what way is this a mapping? Are we mapping keyboard button ids to game controller IDs or something? I guess I don't understand the point of this struct then. I thought it was to handle simultaneous input from a keyboard and game controller.

I'll try to explain this in the easiest way possible

`KeyMapping` -> `ControlMapping` (allows keyboard button ID and  controller button ID stored in `ControlBind`) -> `ButtonMapping` (`ControlBind` renamed to `ButtonInfo`)

Wilyicaro: I honestly don't feel comfortable discussing such silly things like naming

Wilyicaro: I say A, you understand B, I start talking about C, you go back to B, and then you understand A

BrentDaMage: We can discuss it on Discord if you'd like. The thing is, naming is incredibly important to allow contributors to easily understand the purpose of something. So I'd like to work out something that is reasonably clear and concise. Part of the issue is that *I* don't understand entirely what the purpose of these classes are, and I'm having difficulty understanding your explanation. Either I'm dumb, or this may be indicative of a structural or simplicity problem.

Wilyicaro: > We can discuss it on Discord if you'd like. The thing is, naming is incredibly important to allow contributors to easily understand the purpose of something. So I'd like to work out something that is reasonably clear and concise. Part of the issue is that _I_ don't understand entirely what the purpose of these classes are, and I'm having difficulty understanding your explanation. Either I'm dumb, or this may be indicative of a structural problem.

I understand that naming is very important, but a general consensus is almost impossible; what's good for you might not be good for me and vice versa. The difference is that I usually only agree (when I don't like it) cuz I can get used to it.

BrentDaMage: Alright, whatever. I too will give up on finding a consensus.

source/client/app/Minecraft.cpp
Wilyicaro: > When we add split-screen, wouldn't we just have a way to assign an input method to each player instead? I guess this works in the mean time, but its use isn't super apparent to an average user.

yes, but the mouse could be annoying if everyone is using controllers

BrentDaMage: Yeah, but couldn't you just lock the mouse to the first player's area?

Wilyicaro: I just removed this option, it wasn't working right anyway

source/client/app/Minecraft.cpp
```diff
@@ -516,7 +517,10 @@ void Minecraft::handleBuildAction(const BuildActionIntention& action)
 
 void Minecraft::tickInput()
 {
-	if (!m_pInputHolder->allowsType(IInputHolder::activeType))
+	if (!platform()->hasGamepad() && useController())
+		setInputType(platform()->isTouchscreen() ? InputType::TOUCHSCREEN : InputType::KEYBOARD);
```
BrentDaMage: Abstract this into a `resetInputType` function.
Something like this:
```cpp
if (platform()->hasGamepad())
    setInputType(InputType::CONTROLLER);
else if (platform()->isTouchscreen())
    setInputType(InputType::TOUCHSCREEN);
else
    setInputType(InputType::KEYBOARD);
```

source/client/app/NinecraftApp.cpp
```diff
@@ -105,9 +105,9 @@ void NinecraftApp::_initInput()
 
 	//If someone has a gamepad connected, certainly they want to use it
 	if (platform()->hasGamepad())
```
BrentDaMage: Use `resetInputType` function. mentioned above

source/client/gui/components/Button.cpp
```diff
@@ -32,7 +32,7 @@ void Button::_renderBgPocket(Minecraft* mc, const MenuPointer& pointer)
 	int iXPos = isSelected() ? 66 : 0;
 	currentShaderColor = isEnabled() ? Color::WHITE : Color::GREY;
 	mc->m_pTextures->loadAndBindTexture("gui/touchgui.png");
-	blit(m_xPos, m_yPos, iXPos, 0, m_width, m_height, 66, 26, &m_materials.ui_textured_and_glcolor);
+	blitNineSlice(*mc->m_pTextures, TextureAtlasSprite(iXPos, 0, 66, 26, "gui/touchgui.png", 256, 256), m_xPos, m_yPos, m_width, m_height, 3, &m_materials.ui_textured_and_glcolor);
```
BrentDaMage, line 35: So you're aware, there is an issue with the texture atlases where resetting `Textures` doesn't regenerate the atlases, and results in missing textures. This causes problems on platforms where we can lose the render context. This primarily affects iOS and Android when apps are suspended and resumed, as well as SDL1 macOS & Direct3D 9 when the window is resized.

source/client/gui/screens/inventory/ContainerScreen.cpp
```diff
@@ -13,7 +13,7 @@ ContainerScreen::ContainerScreen(ContainerMenu* menu) :
     m_topPos(0),
     m_timeSlotDragged(0)
 {
-    m_themeSelection = UI_UNIVERSAL;
+    m_screenType = SCREEN_UNIVERSAL;
```
BrentDaMage, line 20: Sorry to only suggest this now, but perhaps we call *this* the Screen profile?
We have the overall UI theme, and each screen has its own screen profile. The downside is that this name doesn't do a great job of conveying the difference between the two. Though on the upside, it's more specific than "screen type". We could also call it the "screen layout". Would like to know your thoughts though.

source/client/gui/screens/CreateWorldScreen_Console.cpp
```diff
@@ -38,15 +40,20 @@ void CreateWorldScreen_Console::_buttonClicked(Button* pButton)
...
 		}
 
-		LevelSettings levelSettings(seed);
+		LevelSettings levelSettings(seed, m_gameMode);
 		m_pMinecraft->selectLevel(levelUniqueName, levelNickname, levelSettings);
 	}
+	else if (pButton->getId() == m_btnGameMode.getId())
+	{
+		m_gameMode = static_cast<GameType>((static_cast<int>(m_gameMode) + 1) % (GAME_TYPES_MAX + 1));
+		m_btnGameMode.setMessage("Game Mode: " + GameTypeConv::GameTypeToNonLocString(m_gameMode));
```
BrentDaMage: Sweet. Since this is an addition, could you localize the "Game Mode:" part? Thanks.

source/client/gui/screens/CreateWorldScreen_Console.cpp
```diff
@@ -67,12 +74,18 @@ void CreateWorldScreen_Console::init()
 	m_textSeed.m_yPos = m_panel.y + 188;
 	_addElement(m_textSeed);
 
+	m_btnGameMode.setMessage("Game Mode: " + GameTypeConv::GameTypeToNonLocString(m_gameMode));
```
BrentDaMage: Same here. Maybe abstract this?

source/client/options/Options.cpp
```diff
@@ -418,188 +417,188 @@ std::vector<std::string> Options::getOptionStrings()
...
-	KM(KM_TOGGLEGUI,    "key.fn.gui",        0x70); // VK_F1
-	KM(KM_SCREENSHOT,   "key.fn.screenshot", 0x71); // VK_F2
-	KM(KM_TOGGLEDEBUG,  "key.fn.debug",      0x72); // VK_F3
-	KM(KM_TOGGLEAO,     "key.fn.ao",         0x73); // VK_F4
-	KM(KM_TOGGLE3RD,    "key.fn.3rd",        0x74); // VK_F5
-	KM(KM_FLY_UP,       "key.fly.up",        'X');
-	KM(KM_FLY_DOWN,     "key.fly.down",      'C');
-	KM(KM_CHAT_CMD,     "key.chat.cmd",      0xBF); // VK_OEM_2
-#undef KM
+#define BM(idx, name, code) m_buttonMappings[idx] = ButtonMapping(name, code)
```
BrentDaMage: Again, very sorry to be so picky about names, but I think it would be much clearer if we called these "actions". We already have the "BuildActionIntention" class, so this would it nicely. That way `ButtonInfo` can be called `ActionInfo`.

source/client/player/input/InputType.hpp
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+class InputType
```
BrentDaMage: Maybe call this `InputMethod`, up to you.

source/client/renderer/Font.cpp
```diff
@@ -168,49 +164,7 @@ void Font::drawWordWrap(const std::vector<std::string>& lines, int x, int y, con
...
-
-		MatrixStack::Ref xmtx = MatrixStack::World.push();
-		xmtx->translate(Vec3(xOff, 0.0f, 0.0f));
-		m_charMeshes[x].render(m_materials.ui_text);
-
-		xOff += m_charWidthFloat[x];
-	}
-
-	currentShaderColor = Color::WHITE;
+	drawSlow(str, x, y, color, bShadow);
```
BrentDaMage, line 170: What happened here?

BrentDaMage (changes requested):

source/client/renderer/Font.cpp
Wilyicaro: It turned out that drawing each baked font char mesh wasn't a good idea at all, and the actual `drawSlow` method isn't slow at all

source/client/gui/screens/inventory/ContainerScreen.cpp
BrentDaMage: On second thought, this name is perfect.

source/client/gui/screens/CreateWorldScreen_Console.cpp
BrentDaMage: Doesn't matter, in master now.

source/client/gui/screens/CreateWorldScreen_Console.cpp
BrentDaMage: Doesn't matter, in master now.

BrentDaMage:
Tested swapping between controller, mouse & keyboard, and touchscreen, all worked.

BrentDaMage (approved):

PR #523, opened 2026-02-26
Add Linux flatpak build + aarch64 artifacts

Un1q32:
Builds for x86_64 and aarch64.
Can also build for i386 but the performance overhead of the flatpak sandbox is so high there that it's not worth it.

### DIFF
--- a/.github/workflows/artifacts.yml
+++ b/.github/workflows/artifacts.yml
@@ -73,64 +73,64 @@ jobs:
           - name: Pack cache
             run: cd ~ && tar cJf ccache.tar.xz .cache/ccache
 
-    flatpak:
-        strategy:
-            fail-fast: false
-            matrix:
-                include:
-                  - name: x86_64 SDL 2 (OGL)
-                    gfx-api: OGL
-                    arch: x86_64
-                    runner: ubuntu-24.04
-                  - name: x86_64 SDL 2 (OGL + Shaders)
-                    gfx-api: OGL_SHADERS
-                    arch: x86_64
-                    runner: ubuntu-24.04
-                  - name: aarch64 SDL 2 (OGL)
-                    gfx-api: OGL
-                    arch: aarch64
-                    runner: ubuntu-24.04-arm
-                  - name: aarch64 SDL 2 (OGL + Shaders)
-                    gfx-api: OGL_SHADERS
-                    arch: aarch64
-                    runner: ubuntu-24.04-arm
-        name: Flatpak (${{ matrix.name }})
-        runs-on: ${{ matrix.runner }}
-        steps:
-          - name: Get Time
-            id: get-time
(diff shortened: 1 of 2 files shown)


.github/workflows/artifacts.yml
```diff
@@ -10,14 +10,24 @@ jobs:
...
             matrix:
                 include:
-                  - name: SDL 2 (OGL)
-                    flags: '-DREMCPE_PLATFORM=sdl2 -DREMCPE_GFX_API=OGL'
-                    packages: 'libsdl2-dev'
-                  - name: SDL 2 (OGL + Shaders)
-                    flags: '-DREMCPE_PLATFORM=sdl2 -DREMCPE_GFX_API=OGL_SHADERS'
-                    packages: 'libsdl2-dev'
+                  - name: x86_64 SDL 2 (OGL)
+                    gfx: OGL
```
BrentDaMage: Could u call this `gfx-api`?

platforms/flatpak/.gitignore
BrentDaMage, line 1: Call the `flatpak` directory `linux`, and rename `build.sh` to `build-flatpak.sh`

BrentDaMage:
I'm not a huge fan of *just* distributing a Flatpak package. I'd rather we provide a basic ZIP as well that gets packaged before the Flatpak is built.

Un1q32:
brought back the old artifact build in addition to the flatpak build, also added aarch64 to the old builds

platforms/linux/io.github.reminecraftpe.desktop
```diff
@@ -0,0 +1,10 @@
+[Desktop Entry]
+Name=ReMinecraftPE
+GenericName=ReMinecraftPE
+Comment=Decompilation of Minecraft Pocket Edition with extra features
```
BrentDaMage: Can you use the description from the GitHub repo?

BrentDaMage (changes requested):
Please disable the Flatpak artifacts for now until we come up with a proper name. I believe we're flying too close to the sun.

Un1q32:
> Please disable the Flatpak artifacts for now until we come up with a proper name. I believe we're flying too close to the sun.

To be clear, we don't publish to flathub.  We just make .flatpak packages.  You still need to download them from the github, not from any app store.

BrentDaMage:
> > Please disable the Flatpak artifacts for now until we come up with a proper name. I believe we're flying too close to the sun.
> 
> To be clear, we don't publish to flathub. We just make .flatpak packages.

I know, even then, I think having that Flatpak metadata file in the repo just isn't a good idea. If we ever do Flatpak, we need to have a better name.

Un1q32:
I personally don't see how its any worse than a tarball, but whatever.

Un1q32:
Flatpak workflows commented

BrentDaMage (approved):

PR #528, opened 2026-03-01
Finished rebrand to NBCraft

BrentDaMage:


platforms/ios/NBCraft-Info.plist
iProgramMC, line 1: This looks like it needs to be renamed `nbcraft-Info.plist`.

Un1q32: fixed in my pr

BrentDaMage: I have a WIP PR I fixed this in as well. It fixes some visual bugs on 360.

PR #531, opened 2026-03-01
Rebrand to NBCraft on Android

iProgramMC:
As part of the ongoing rebrand.

### DIFF
--- a/platforms/android/project/app/build.gradle
+++ b/platforms/android/project/app/build.gradle
@@ -54,7 +54,7 @@ android {
         // testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
         externalNativeBuild {
             cmake {
-                arguments '-DANDROID_PLATFORM=android-21', '-DANDROID_STL=c++_static', '-DREMCPE_PLATFORM=android', '-DREMCPE_AUDIO_LIBRARY=opensl'
+                arguments '-DANDROID_PLATFORM=android-21', '-DANDROID_STL=c++_static', '-DNBC_PLATFORM=android', '-DNBC_AUDIO_LIBRARY=opensl', '-DNBC_GFX_API=OGL'
                 abiFilters 'armeabi-v7a', 'arm64-v8a', 'x86', 'x86_64'
             }
         }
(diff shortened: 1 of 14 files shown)


platforms/android/project/app/build.gradle
```diff
@@ -3,7 +3,7 @@ plugins {
 }
 
 android {
-    namespace 'com.reminecraftpe'
+    namespace 'com.nbcraft'
```
Un1q32, line 6: this should be org.nbcraft

iProgramMC: Was `com.reminecraftpe` for like 2 years and nobody complained.

iProgramMC:
Please do not pull yet as I don't know if this works.

platforms/android/project/app/build.gradle
Un1q32: we're using org.nbcraft everywhere, and org.nbcraft.nbcraft instead of io.github.reminecraftpe

Un1q32: just let me do it cuz i was already working on this, just gotta fix macos CI

iProgramMC: If we use `org.nbcraft` for both, then there will be package confusion between the Android native activity-based version and the SDL2 version.

iProgramMC:
Android native turns out to be broken in more ways than one

iProgramMC:
@Un1q32 you can take care of the SDL rename, I'll undo my own commits for SDL Android and I'll rename this PR to "fix android native + rebrand" or something

Un1q32:
To fix android native you would have to fix the OpenSL backend

iProgramMC:
will be finished in another PR.

platforms/android/project/app/build.gradle
BrentDaMage: One should be called `org.nbcraft.nbcraft-sdl2` and the other `org.nbcraft.nbcraft-native` or whatever. Perhaps native just gets `org.nbcraft.nbcraft`. I have a similar naming for the Xcode and VS clients. `NBCraftClient.SDL2`, `NBCraftClient.Win32`, `NBCraftClient.iOS`, etc.

PR #533, opened 2026-03-01
Fix Android native build

iProgramMC:
NOTE: Seems like globally initialized objects are broken, because only making `AppPlatform_android` a pointer in main.cpp fixed a crash issue regarding `AppPlatform::m_singleton` seemingly being cleared.

For some reason, it seems like libnbcraft.so is loaded twice? Didn't happen in the last PR.
image


iProgramMC:
This PR is now ready. I can now compile an android APK for NBCraft with this project.

PR #536, opened 2026-03-02
Windows XP build script

Un1q32:


.github/workflows/publish.yml
```diff
@@ -202,7 +202,7 @@ jobs:
               releaseDirectory: ${{ matrix.directory }}/app/build/outputs/apk/release
               signingKeyBase64: ${{ secrets.ANDROID_KEYSTORE }}
               keyStorePassword: ${{ secrets.ANDROID_KEYSTORE_PASSWORD }}
-              alias: NBCraft
+              alias: remcpe
```
BrentDaMage, line 205: fix

platforms/windows/.gitignore
```diff
@@ -0,0 +1,2 @@
+/build
+/NBCraftClient.Win32.configured.rc
```
BrentDaMage: just ignore all `*.configured.rc`

platforms/windows/gcc.diff
BrentDaMage, line 1: What is this enabling us to do? Build for Windows 2000?

.github/workflows/publish.yml
Un1q32: We would have to generate a new signature for android signing, the alias is part of the signature.

platforms/windows/gcc.diff
Un1q32: Yea it removes uses of APIs only present in newer windows versions.  It's just a reindexed partial version of https://github.com/DiscordMessenger/dm/blob/master/doc/pentium-toolchain/gcc.diff, made by iProgramMC.

BrentDaMage: Can you note that in a comment at the top of the diff somehow?

PR #555, opened 2026-03-07
Remove buffer limits in Util::vformat

### LINKED ISSUE #550: Remove hard-coded buffer limit in Util::vformat

Un1q32:
closes #550

This is a lot less efficient since we run snprintf twice, there is no other way around this.
@bluepilledgreat Do we even want this?  If 1024 isn't enough we can just make it 2048 or 4096 instead, running printf twice is inefficient.


bluepilledgreat:
This is slower than other implementations atm

PR #559, opened 2026-03-08
dynamic xinput

Un1q32:
Dynamically load xinput so the game can work on Windows 2000.

### DIFF
--- a/platforms/input/xinput/XInput.hpp
+++ b/platforms/input/xinput/XInput.hpp
@@ -21,7 +21,7 @@ public:
 #ifdef _XBOX
         GetState = ::XInputGetState;
 #else
-        HMODULE module =      LoadLibrary("XINPUT1_4.DLL");
+        HMODULE module      = LoadLibrary("XINPUT1_4.DLL");
         if (!module) module = LoadLibrary("XINPUT1_3.DLL");
         if (!module) module = LoadLibrary("XINPUT1_2.DLL");
         if (!module) module = LoadLibrary("XINPUT1_1.DLL");


platforms/input/xinput/GameControllerHandler_xinput.cpp
```diff
@@ -16,6 +16,13 @@
 GameControllerHandler_xinput::GameControllerHandler_xinput()
 	: GameControllerHandler()
 {
+    getState = nullptr;
+    HMODULE module = LoadLibraryA("XINPUT1_3.dll");
```
BrentDaMage: woah woah woah, we can't be doing this on the Xbox 360, which shares this class. Can you abstract the new stuff into a private `_initBindings` function (kinda like we have in mce::Platform::OGL)? Then, only compile the contents of the function if `MC_PLATFORM_WINPC` is true.

platforms/input/xinput/GameControllerHandler_xinput.cpp
```diff
@@ -84,10 +91,16 @@ void GameControllerHandler_xinput::_processMotion(GameController::ID controllerI
 
 void GameControllerHandler_xinput::refresh()
 {
+    if (!getState)
```
BrentDaMage: We shouldn't even need to write to an array that theoretically will never change state.

BrentDaMage:
Also, perhaps we put the bindings in some sort of static class inside of the `Input.XInput` project instead? I'm not a huge fan of having the function bindings/pointers inside of the `GameControllerHandler`.

platforms/input/xinput/GameControllerHandler_xinput.cpp
Un1q32: Don't we need to set STATE_DISCONNECTED so hasGamepad will return false?  Otherwise the memory would be uninitialized

BrentDaMage: I'm not telling you to never initialize the array. Just write to it once on initialization, then never again. Afaik, `refresh` is called every frame.

Un1q32:
> Also, perhaps we put the bindings in some sort of static class inside of the `Input.XInput` project instead? I'm not a huge fan of having the function bindings/pointers inside of the `GameControllerHandler`.

Not sure what this means, like create a whole new source file and class to abstract one xinput function?  What would this accomplish?

platforms/input/xinput/GameControllerHandler_xinput.cpp
Un1q32: The OGL InitBindings function gets called multiple times in different source files, better to inline in this case I think as it's only being called in one place ever.  Speaking of, that _initButtonMap function isn't called anywhere else either, any reason not to inline that?

BrentDaMage:
> > Also, perhaps we put the bindings in some sort of static class inside of the `Input.XInput` project instead? I'm not a huge fan of having the function bindings/pointers inside of the `GameControllerHandler`.
> 
> Not sure what this means, like create a whole new source file and class to abstract one xinput function? What would this accomplish?

Correct. This would ensure that the XInput function bindings remain separate from the `GameControllerHandler` itself. That way, we can have the ugly macro checks separate from the actual logic for handling game controller input. Additionally, this class should be static.

platforms/input/xinput/GameControllerHandler_xinput.cpp
BrentDaMage: There's really no substantial benefit to inlining the function, since it's only called once during the game's initialization phase. Additionally, in order to inline it, we'd need to move it to the header, which would just clutter the header.

Un1q32:
> Correct. This would ensure that the XInput function bindings remain separate from the `GameControllerHandler` itself. That way, we can have the ugly macro checks separate from the actual logic for handling game controller input. Additionally, this class should be static.

Wouldn't this add additional runtime and memory overhead?  Doesn't seem worth it to move one macro check somewhere else.  Plus it means someone reading the code for the first time has to read yet another file to understand it.

BrentDaMage:
> > Correct. This would ensure that the XInput function bindings remain separate from the `GameControllerHandler` itself. That way, we can have the ugly macro checks separate from the actual logic for handling game controller input. Additionally, this class should be static.
> 
> Wouldn't this add additional runtime and memory overhead? Doesn't seem worth it to move one macro check somewhere else. Plus it means someone reading the code for the first time has to read yet another file to understand it.

There would be no additional memory overhead. You're taking a member variable from within a singleton class, and moving it to a static class. Combining function bindings and controller handling logic into one class is stupid imo. Having these two parts separated makes things cleaner and easier to understand. They should be separate, like they are in OGL. If you'd like, you could make a static boolean as well to track whether or not the functions have been bound, then just call the static class's `InitBindings` function in the constructor of the `GameControllerHandler`.

Un1q32:
Wait I was thinking objects

platforms/input/xinput/GameControllerHandler_xinput.cpp
Un1q32: > There's really no substantial benefit to inlining the function, since it's only called once during the game's initialization phase. Additionally, in order to inline it, we'd need to move it to the header, which would just clutter the header.

We wouldn't move it to the header we would just copy the function contents to the initializer and remove the function entirely from the header and source file.

BrentDaMage: I like what we have now, I think it's more organized, and the performance impact is insanely negligible. One function call on init.

platforms/input/xinput/XInput.cpp
```diff
@@ -0,0 +1,3 @@
+#include "XInput.hpp"
+
+DWORD (WINAPI *XInput::XInputGetState)(DWORD, XINPUT_STATE *) = nullptr;
```
BrentDaMage: Name this `XInput::GetState`. Makes more sense, no?

platforms/input/xinput/XInput.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+    static DWORD (WINAPI *XInputGetState)(DWORD, XINPUT_STATE *);
+
+    static void init(void)
+    {
+#ifdef _XBOX
+        XInputGetState = ::XInputGetState;
+#else
+        XInputGetState = nullptr;
+        HMODULE module = LoadLibraryA("xinput1_3.dll");
```
BrentDaMage: Could we try binding `xinput_1_3.dll`, then `xinput_1_2.dll` and so on? Is the `xinput.h` header sensitive on this, or are we good? If this works, this could certainly bring the minimum requirement for controller support on Window down a bit.

BrentDaMage (changes requested):

platforms/input/xinput/XInput.hpp
Un1q32: We only need the one function, is the function signature the same in all the dll versions?

BrentDaMage: I'm not sure. I think our bigger concern is the structs. The documentation doesn't really say that anything *changed*, just that new stuff was added over time.
Can you have it try to load to the following DLLs in the following order?
```
XINPUT1_4.DLL
XINPUT1_3.DLL
XINPUT1_2.DLL
XINPUT1_1.DLL
XINPUT9_1_0.DLL (this is actually version 1.0, just prefixed with 9 due to being released with DirectX 9)
```
Thanks

BrentDaMage: Also, you should not have to call "LoadLibraryA" directly, just do "LoadLibrary", which is a macro that will pick the right version based on whether or not we have unicode support enabled in MSVC (which we don't).

platforms/input/xinput/XInput.hpp
```diff
@@ -0,0 +1,33 @@
...
+public:
+
+    static DWORD (WINAPI *XInputGetState)(DWORD, XINPUT_STATE *);
+
+    static void init(void)
+    {
+#ifdef _XBOX
+        XInputGetState = ::XInputGetState;
+#else
+        XInputGetState = nullptr;
```
BrentDaMage: You're already setting this to nullptr in the .ccp file.

platforms/input/xinput/XInput.hpp
Un1q32: https://learn.microsoft.com/en-us/windows/win32/xinput/xinput-versions
Screenshot 2026-03-14 at 12 17 21 PM

Microsoft seems to want us to just use the oldest version only in this kind of case, reading the page going over the differences it seems like there's no downside.

BrentDaMage: Right, but what if they stop shipping the older DLLs? We might as well try to bind to the newer ones.

Un1q32: > Right, but what if they stop shipping the older DLLs? We might as well try to bind to the newer ones.

They will literally never but ok.

BrentDaMage: It also doesn't explicitly say that XIinput 9.1.0 ships with Windows 10. Not to mention that none of this mentions Windows 11. It's entirely possible that they don't ship these older DLLs with Windows in the future, and instead require you to install a redistributable package.

platforms/input/xinput/XInput.hpp
```diff
@@ -0,0 +1,40 @@
...
+    static DWORD (WINAPI *GetState)(DWORD, XINPUT_STATE *);
+
+    static void init(void)
+    {
+#ifdef _XBOX
+        GetState = ::XInputGetState;
+#else
+        HMODULE module = LoadLibrary("XINPUT1_4.DLL");
+        if (!module)
+            module = LoadLibrary("XINPUT1_3.DLL");
```
BrentDaMage: Could u make the if check and the contents one line?

platforms/input/xinput/XInput.hpp
```diff
@@ -0,0 +1,36 @@
...
+public:
+
+    static DWORD (WINAPI *GetState)(DWORD, XINPUT_STATE *);
+
+    static void init(void)
+    {
+#ifdef _XBOX
+        GetState = ::XInputGetState;
+#else
+        HMODULE module = LoadLibrary("XINPUT1_4.DLL");
```
BrentDaMage: Put spaces to the left of the equals to make each LoadLibrary call horizontally aligned with one another.

platforms/input/xinput/XInput.hpp
```diff
@@ -0,0 +1,36 @@
...
+public:
+
+    static DWORD (WINAPI *GetState)(DWORD, XINPUT_STATE *);
+
+    static void init(void)
+    {
+#ifdef _XBOX
+        GetState = ::XInputGetState;
+#else
+        HMODULE module =      LoadLibrary("XINPUT1_4.DLL");
```
BrentDaMage: Can u make the equals aligned?

BrentDaMage (approved):

PR #562, opened 2026-03-09
update zlib to 1.3.2 and static link zlib

Un1q32:
Apparently they fixed some RCE.

Should we always static link zlib?  Some platforms where we dynamic link, like old iOS and macOS versions, have very old versions with multiple RCE vulnerabilities.


BrentDaMage:
Given the circumstances, I'd say always static link if possible.

thirdparty/zlib/gzguts.h
```diff
@@ -36,21 +48,15 @@
 
 #if defined(__TURBOC__) || defined(_MSC_VER) || defined(_WIN32)
 #  include <io.h>
+#  include <sys/stat.h>
 #else
 #  include <unistd.h>
 #endif
 
-#if defined(_WIN32) && !defined(_XBOX)
+#if defined(_WIN32) && !defined(WIDECHAR) && !defined(_XBOX)
```
BrentDaMage: Can you flip the WIDECHAR and _XBOX order here?

PR #478, opened 2026-02-15
Tile Entity Logic

samdotjpg:
Okay I'll update this with more information about the PR and its features/changes when it's nearing completion. Right now I'm just documenting my progress and what's left to do.

- [ ] Chests
- - [x] Singleplayer functionality
- - [x] Fix UI from console 
- - [ ] ~~Save compatibility check~~ <- 0.4.0? where should we take this from? unneeded for now if goal is 0.3.3
- - [ ] Multiplayer support <- Once ContainerListener is fixed, this should work instantly. This isn't a TileEntity specific issue, but we need to be able to:
  - - Open / close chest
  - - See container items
  - - Move items in container
- [ ] Furnaces
- - [x] Singleplayer functionality
- - [x] Save compatibility check
- - [ ] Multiplayer support <- Same deal as chests
- [x] Note Blocks
- - [x] Singleplayer functionality
- - [ ] Multiplayer support <- This will naturally work when TileEvent is fixed and players don't place a block/item when interacting with containers/crafting tables/etc.

misc:
- entities stored as maps instead of vectors
  - tile entities will **need** to be migrated to this once we have an unordered map so we can hash them based on their coordinates 
- riding entity logic
- changed "looking at" entities to fetch an entity by ID, which closes #342 
  - can someone please make (or choose an existing) o(1) hashmap? we have so many inefficient map/vector functions
- charcoal
- facing:: helpers
- touchscreen controller support
- sheep are no longer gross on gles
- sand/gravel drop items now, closes #447
- snow tiles drop snowballs, closes #449

Right now even interacting with tile entities in multiplayer causes your buddy's client watch you place whatever is in your hand.

PR #577, opened 2026-03-31
Add Furnace block with state persistence

DomAtTheShack:
- Implement Furnace block with crafting and smelting mechanics.
- Add NBT-based chunk saving via LevelChunk to persist furnace state and menu data.
- Establish baseline data structures for future interactable entities (e.g., chests).

### DIFF
--- a/game/assets/.gitignore
+++ b/game/assets/.gitignore
@@ -101,3 +101,4 @@
 !/patches/p_info_updategame1.png
 !/patches/p_info_updategame2.png
 !/patches/p_slot_more.png
+/resource_packs/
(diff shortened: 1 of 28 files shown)


Un1q32:
If you wanna do furnaces and tile entities it should be based on #478, we wanna keep compatibility with the original PE network protocol for now. AFAIK it was nearly finished but the person doing it left due to personal issues.

BrentDaMage:
> If you wanna do furnaces and tile entities it should be based on #478, we wanna keep compatibility with the original PE network protocol for now. AFAIK it was nearly finished but the person doing it left due to personal issues.

This ^
Additionally, it seems like this PR is mainly vibecoded. At the end of the day, I would prefer the pre-existing man-made PR, as opposed to this, since it's likely had much more thought put into it. There was also a fair bit of work done in the previous PR to fix bugs with container replication, which I don't believe this current PR has. Apologies for the confusion. Feel free to open another PR, should you choose to base it on #478.

PR #578, opened 2026-04-02
Windows build script docs

Un1q32:
Also makes the build script work on macOS, this is the only way to build for 9x so it should be documented.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,14 +141,14 @@ You will need to install the following dependencies
 Then run
 
 ```sh
-# for 64 bit machines
+# to build for 64 bit machines
 export ARCH=x86_64
-# for pentium pro or newer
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -114,6 +114,59 @@ This fetches the project's necessary dependencies.
...
+
+- Xcode command line tools
+- `cmake` (CMake)
+- `wget` (Wget)
+- `gmp mpfr mpc` (GCC dependencies)
+
+Then run
+
+```sh
+# for 64 bit machines
```
BrentDaMage: make it clear that this is for **targeting** those machines

BrentDaMage (changes requested):

BrentDaMage (approved):

PR #580, opened 2026-04-08
macOS .app + fix cache for windows and mac.

Un1q32:
macOS builds are roughly 2x as fast
Windows builds are roughly 3x as fast

macOS will need 1 slow build before speeding up because it was a bug in how the cache was built.

Closes #566


platforms/macos/Info.plist
```diff
@@ -0,0 +1,22 @@
...
+<!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>CFBundleDevelopmentRegion</key>
+	<string>English</string>
+	<key>CFBundleExecutable</key>
+	<string>nbcraft</string>
+	<key>CFBundleIconFile</key>
+	<string>icon.icns</string>
+	<key>CFBundleIdentifier</key>
```
BrentDaMage, line 11: This should be `org.nbcraft.NBCraftClient.MacOS`

BrentDaMage (changes requested):
Make bundle identifier match our convention for the iOS client.

Un1q32:
done

BrentDaMage (approved):

PR #584, opened 2026-04-11
Fix building for OpenBSD and FreeBSD

Un1q32:


BrentDaMage:
We support OpenBSD?

BrentDaMage (approved):

Un1q32:
After this patch yeah.

BrentDaMage:
> After this patch yeah.

Should we add it to the list?

Un1q32:
Sure, lemme try building on FreeBSD too cuz I have a feeling that will also just work.

Un1q32:
Basically anything with complete enough POSIX support, a C++98 compliant compiler and runtime, SDL2, and OpenGL 1.1 support can build and run the game without too much trouble.

BrentDaMage:
> Basically anything with complete enough POSIX support, a C++98 compliant compiler and runtime, SDL2, and OpenGL 1.1 support can build and run the game without too much trouble.

Maybe we put next to Linux `(or anything with POSIX, SDL, and OpenGL support)`?

Un1q32:
Yea, I've thought about trying to add a new POSIX backend that uses X11 directly so you wouldn't even need SDL, then it would with hardly any work on virtually any unix-like system new enough to support OpenGL, but that's probably too ambitious for me since I've never written anything using X11 before.

BrentDaMage (approved):

PR #588, opened 2026-04-13
GNU Makefile build

Un1q32:
I've been trying to run NBCraft on weird old unix systems lately and the biggest obstacle has been CMake, which requires a modern C++ compiler/libraries as well as libuv (a library that does not care about portability at all).  A GNU makefile makes it ***much*** easier to get running on old systems, especially since our rigid adherence to C++98 has made it possible to build with very old compilers.
I've tested the build on Debian 4.0 with GCC 3.4 and it works, GCC 3.3 and below do not work.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ CFLAGS := -O2 -DNDEBUG
 CXXFLAGS := -O2 -DNDEBUG
 
 DEFINES := -DHANDLE_CHARS_SEPARATELY -DRAPIDJSON_NO_THREAD_LOCAL -DSTBI_NO_THREAD_LOCALS
-INCLUDES := -I. -Isource -Ithirdparty -Ithirdparty/zlib -Ithirdparty/raknet -Ithirdparty/rapidjson -Ithirdparty/stb_image/include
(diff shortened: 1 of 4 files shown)


TimofeyLednev:
Maybe I should try porting this to BeOS for fun? Everything seems to be available for porting: SDL 1.2.5, OpenGL 1.1 and apparently the Makefile is coming soon

Un1q32:
> Maybe I should try porting this to BeOS for fun? Everything seems to be available for porting: SDL 1.2.5, OpenGL 1.1 and apparently the Makefile is coming soon

Give it a shot!  The whole point of the makefile is to make it easier to build for weird stuff.  I tested builds on old Debian versions and it seems to work fine with GCC 3.4 or later.

TimofeyLednev:
There was a non-official port of GCC 3.4.0 on BeOS, but it's very bad, so I'll use what's built into the system, namely gcc 2.95.3, although I don't know if I'll be able to compile the game, but I'll try it

Makefile
BrentDaMage, line 1: I will try to make changes to this as needed, but it will be your job to maintain it.

BrentDaMage (approved):

PR #590, opened 2026-04-19
Add option to dynamically use OpenGL symbols on unix-like systems.

Un1q32:
It's disabled by default on non-windows because it would break static builds, which would be bad for static-only platforms like the switch.  It also wouldn't help much for Linux artifact builds, since those are linked against either the flatpak runtime, which is guaranteed to have a new enough mesa, or a relatively new glibc which is much much newer than the last version of mesa to not have every required function.

Also includes major improvements to the Makefile build, and fixes for extremely old systems, such as systems with very old mesa headers, and Mac OS X 10.3.

Also adds support for GCC versions down to 2.96 (Redhat's version of GCC 3.0 beta)

Also updates the GCC version used in the Windows and macOS artifacts to 16.1.0, important because GCC 15.2.0 can't be built by 16.1.0 for some reason, so building the artifact is currently impossible on an updated Arch Linux install.


source/common/Util.cpp
```diff
@@ -154,6 +154,8 @@ std::string Util::toString(const wchar_t* str)
 	return result;
 }
 
+#ifndef NBC_NO_WIDECHAR
```
BrentDaMage: Make this `MC_NO_WIDECHAR`. All CMake variables should be prefixed with `NBC_`, all C/C++ macros should be prefixed with `MC_`.

source/renderer/hal/ogl/API_OGL.cpp
```diff
@@ -219,7 +219,7 @@ bool gl::supportsServerBuffers()
         const std::string& glExtensions = gl::getOpenGLExtensions();
         // OpenGL ES || OpenGL 1.5+
         if ((glVersion.gles || glExtensions.find("GL_ARB_vertex_buffer_object") || (glVersion.major > 1 || glVersion.minor >= 5))
-#ifdef _WIN32
+#if DYNAMIC_OPENGL
```
BrentDaMage: Name this `MC_DYNAMIC_GL`

thirdparty/GL/GL.hpp
```diff
@@ -88,6 +103,15 @@
 	// use our macro for glOrtho
 #endif
 
+#if DYNAMIC_OPENGL && !defined(_WIN32) && !defined(__DREAMCAST__)
```
BrentDaMage: Can you make a `MC_PLATFORM_UNIX` macro in `compat/PlatformDefinitions.h`?
Define it as `!defined(_WIN32) && !defined(__DREAMCAST__)`, that way you can use it above and on other files as needed.

thirdparty/raknet/RakNetSocket2.cpp
```diff
@@ -35,9 +35,6 @@ using namespace RakNet;
 #include <sys/types.h>
 #include <lwip/sockets.h>
 #else
-#if (!defined(ANDROID) && !defined(__DREAMCAST__) && !defined(__SWITCH__))
-#include <ifaddrs.h>
```
BrentDaMage, line 39: Had this for libXenon compatibility.

BrentDaMage (changes requested):

thirdparty/GL/GL.hpp
Un1q32: Now that I think about it, what macro does the dreamcast KalistOS SDK use?  Cuz thats definitely a unix-like and not windows.  I think all this dreamcast code was written like it is with the assumption we would be using the official Windows CE based SDK, but if we ever did a dreamcast port I would think the unix-like homebrew SDK would make much more sense.

thirdparty/raknet/RakNetSocket2.cpp
Un1q32: Does libXenon have a macro I can use so it's only included there?

compat/PlatformDefinitions.h
```diff
@@ -100,3 +100,9 @@
 #else
 #define MC_PLATFORM_CONSOLE 0
 #endif
+
+#if !defined(_WIN32)
```
BrentDaMage: Add the comment please. `Bell Labs - UNIX`, also move the platform definition above the categories.

BrentDaMage (changes requested):

thirdparty/raknet/RakNetSocket2.cpp
BrentDaMage: `MC_SDK_LIBXENON` in `compat/ToolsetDefinitions.h`

compat/limits.hpp
```diff
@@ -0,0 +1,103 @@
+#pragma once
+
+#if !defined(__GNUC__) || (__GNUC__ >= 3)
+
+#include <limits>
+
+#else
+
+/* GCC before 3.0 didn't have <limits> */
```
BrentDaMage: could you just move everything under this comment to its own header file inside of compat? maybe call it `Limits_Legacy.hpp`?
then, just under the comment, include the header.

compat/Limits_legacy.hpp
BrentDaMage, line 1: Rename file to `Limits.hpp` to match casing of other files in compat.

compat/PlatformDefinitions.h
```diff
@@ -80,6 +80,13 @@
 #define MC_PLATFORM_WINPC 0
 #endif
 
+/* Bell Labs - UNIX (and simalar systems) */
+#if !defined(_WIN32)
```
BrentDaMage, line 84: be aware of `defined(__unix__) || defined(unix)`, though I don't know how many UNIX platforms actually define this

source/client/sound/SoundData.cpp
```diff
@@ -16,6 +16,8 @@
 #include "client/resources/ResourcePackManager.hpp"
 #include "client/resources/ResourcePackRepository.hpp"
 
+#include <stdlib.h>
```
BrentDaMage: this should be at the top of the file, above our headers.

source/common/Util.hpp
```diff
@@ -83,9 +85,9 @@ class Util
 		int32_t result = 0;
 
 		const size_t size = str.size();
-		for (size_t i = 0; i < size; i++)
+		for (size_t i = 0; i < size; ++i)
```
BrentDaMage, line 88: thank you

thirdparty/GL/GL.hpp
```diff
@@ -88,6 +103,15 @@
 	// use our macro for glOrtho
 #endif
 
+#if MC_DYNAMIC_GL && MC_PLATFORM_UNIX
+// The Mesa headers included with Red Hat Linux 9.0 (and possibly other systems) don't provide this
+// typedef in the main gl.h. It is provided in glext.h, but that doesn't declare it because it won't
+// do it if GL_VERSION_1_3 is defined and gl.h does still define GL_VERSION_1_3. We get around this
+// by just providing the typedefs ourselves here and using a preprocessor hack to avoid conflicts.
+#define PFNGLACTIVETEXTUREPROC __NBC_PFNGLACTIVETEXTUREPROC
```
BrentDaMage, line 111: If something is defined on the C/C++ side, it should be prefixed with `MC_`, if it's in CMake, use `NBC_`.

thirdparty/GL/glext.h
```diff
@@ -85,10 +85,14 @@ extern "C" {
 #define GL_TEXTURE_MAX_LEVEL              0x813D
 #define GL_SMOOTH_POINT_SIZE_RANGE        0x0B12
 #define GL_SMOOTH_POINT_SIZE_GRANULARITY  0x0B13
-#define GL_SMOOTH_LINE_WIDTH_RANGE        0x0B22
 #define GL_SMOOTH_LINE_WIDTH_GRANULARITY  0x0B23
 #define GL_ALIASED_LINE_WIDTH_RANGE       0x846E
 #endif
+// This used to be above but some old mesa headers
+// define GL_VERSION_1_2 but not this for some reason.
```
BrentDaMage, line 92: might be unsupported? but making calls with this (even though we don't) shouldn't make anything explode.

thirdparty/raknet/SocketLayer.cpp
```diff
@@ -203,15 +203,11 @@ void SocketLayer::SetSocketOptions( __UDPSOCKET__ listenSocket, bool blockingSoc
...
+#if (!defined(__DREAMCAST__) && !defined(MC_NO_NETWORKING) && !defined(__sun__))
 	RakNet::RakString netMaskString;
 	RakNet::RakString ipString;
 
-
-
-
-
-#if   defined(WINDOWS_STORE_RT) || defined(_XBOX) || defined(XENON)
+#if defined(WINDOWS_STORE_RT) || defined(_XBOX) || defined(XENON)
```
BrentDaMage: Replace `defined(_XBOX) || defined(XENON)` with `MC_PLATFORM_XBOX`

thirdparty/rapidjson/document.h
```diff
@@ -51,6 +51,10 @@ RAPIDJSON_DIAG_OFF(effc++)
 #undef GetObject
 #endif
 
+#if defined(__GNUC__) && (((__GNUC__ == 3) && (__GNUC_MINOR__ == 0)) || (__GNUC__ < 3))
+#define RAPIDJSON_NOMEMBERITERATORCLASS
```
BrentDaMage, line 55: This is supposed to be defined by the compiler/toolset or whatever. Not in the code, though I don't really care.

thirdparty/rapidjson/document.h
```diff
@@ -498,15 +502,14 @@ inline GenericStringRef<CharType> StringRef(const std::basic_string<CharType>& s
 // GenericValue type traits
 namespace internal {
 
-template <typename T, typename Encoding = void, typename Allocator = void>
-struct IsGenericValueImpl : FalseType {};
+template <typename T>
+struct IsGenericValue : FalseType {};
 
-// select candidates according to nested encoding and allocator types
```
BrentDaMage, line 504: keep comment

thirdparty/rapidjson/document.h
```diff
@@ -498,15 +502,14 @@ inline GenericStringRef<CharType> StringRef(const std::basic_string<CharType>& s
...
-template <typename T> struct IsGenericValueImpl<T, typename Void<typename T::EncodingType>::Type, typename Void<typename T::AllocatorType>::Type>
-    : IsBaseOf<GenericValue<typename T::EncodingType, typename T::AllocatorType>, T>::Type {};
+template <typename Encoding, typename Allocator>
+struct IsGenericValue< GenericValue<Encoding, Allocator> > : TrueType {};
 
-// helper to match arbitrary GenericValue instantiations, including derived classes
```
BrentDaMage, line 508: keep comment

BrentDaMage (changes requested):

compat/Limits.hpp
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#if !defined(__GNUC__) || (__GNUC__ >= 3)
```
BrentDaMage: swap the order of the if statement here, that way we're not having to read inverted logic

BrentDaMage (changes requested):
one minor thing

BrentDaMage (approved):

PR #602, opened 2026-05-10
CRTDLL support

Un1q32:
Adds support for building with crtdll instead of msvcrt
This isn't enabled in the build script yet because it doesn't allow any new OS to work because of dsound.dll, but it *can* run on NT 3.51 if you build with a dummy audio library.


Un1q32:
Will undraft once I get a setup to test if it actually starts in NT 3.51

Un1q32:
It won't launch because apperently microsoft's OpenGL stack on NT 3.51 was OpenGL 1.0 instead of 1.1 and doesn't have some stuff like glBindTexture.

Un1q32:
Works with some random software mesa dll, so good enough.

source/client/app/AppPlatform.cpp
```diff
@@ -63,6 +63,48 @@ void AppPlatform::finish()
 
 }
 
+#ifdef __CRTDLL__
```
BrentDaMage: Don't throw this Windows-specific shit in here. Maybe put it in `Utils.hpp`?

BrentDaMage (changes requested):

source/client/app/AppPlatform.cpp
Un1q32: > Don't throw this Windows-specific shit in here. Maybe put it in `Utils.hpp`?

It has static global variables so if I put it in a header and that header was included in more than one file there would be desync, and that matters in this case since gmtime_s is supposed to be thread safe.

Un1q32: I could add a new file, but this seems like a small thing to add a new file for, plus that would prolly hurt performance over inlining it.

BrentDaMage: > > Don't throw this Windows-specific shit in here. Maybe put it in `Utils.hpp`?
> 
> It has static global variables so if I put it in a header and that header was included in more than one file there would be desync, and that matters in this case since gmtime_s is supposed to be thread safe.

Declare the functions in Utils.hpp, and define them in Utils.cpp.

source/common/Utils.cpp
```diff
@@ -50,6 +50,41 @@ int g_TimeSecondsOnInit = 0;
...
+    tmp = gmtime(timer);
+    if (tmp)
+        *out = *tmp;
+
+    LeaveCriticalSection(&g_gmtime_s_lock);
+
+    return tmp ? 0 : EINVAL;
+}
+
+#endif
```
BrentDaMage: add a comment after the #endif, ex: `#endif // defined(__CRTDLL__)`

source/common/Utils.hpp
```diff
@@ -35,6 +35,13 @@
 
 #ifdef _WIN32
 
+#ifdef __CRTDLL__
+
+#define gmtime_s __nbc_gmtime_s
+errno_t gmtime_s(struct tm* out, const time_t* timer);
+
+#endif
```
BrentDaMage: same here

BrentDaMage (changes requested):
a couple minor things, other than that, looks good.

BrentDaMage (approved):

PR #609, opened 2026-05-12
Fix renderer warnings

bluepilledgreat:
- Fixed the renderer using `unsigned int` instead of `size_t`. Even though they're both technically the same, MSVC does not like mismatching typedefs and their actual types.
- Fixed `_renderStateTypeMap` being defined twice

### DIFF
--- a/source/renderer/ConstantBufferMetaData.cpp
+++ b/source/renderer/ConstantBufferMetaData.cpp
@@ -41,13 +41,13 @@ const UniformMetaData* ConstantBufferMetaData::getUniformMetaData(const std::str
     return nullptr;
 }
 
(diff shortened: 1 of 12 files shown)


Un1q32:
size_t is not the same as unsigned int on 64 bit anything

bluepilledgreat:
> size_t is not the same as unsigned int on 64 bit anything

was thinking of 32-bit platforms when writing that, whoops

BrentDaMage (approved):
Code looks good, however I'd like to test with an ancient version of Xcode first on my Power Mac G5, since I had to make a lot of changes for the JsonParsers initially for that exact reason.

BrentDaMage:
Code builds fine on the G5.

PR #611, opened 2026-05-12
Add commit info to builds

bluepilledgreat:
none of this is tested


Un1q32:
For the platform build scripts add a fallback where if the environment variables aren't set it first tries to set them by running the right git commands and if that fails it falls back to something like "unknown"

PR #621, opened 2026-06-06
Add scroll wheel support to SDL 1.2 + Makefile improvements

Un1q32:
Makefile changes:
- SDL library dependencies are determined by running pkg-config now, which is the only proper way to do this.  Systems without pkg-config can manually specify the correct libraries with the SDL1_LIBS or SDL2_LIBS variables.
- We now use the proper GCC flags to handle the dependencies of modified header files, instead of doing a full rebuild on every header change.  Setting the DISABLE_MMD variable will go back to the old behavior, useful for people trying to use compilers that don't support the required flags.

SDL 1.2 changes:
- Added scroll wheel support.

### DIFF
--- a/platforms/sdl/sdl1/main.cpp
+++ b/platforms/sdl/sdl1/main.cpp
@@ -152,6 +152,7 @@ static void handle_events()
         {
             case SDL_KEYDOWN:
             case SDL_KEYUP:
+            {
                 if (event.type == SDL_KEYDOWN && event.key.keysym.unicode > 0)
                 {
                     if ((event.key.keysym.unicode & 0xFF80) == 0)
@@ -167,8 +168,10 @@ static void handle_events()
 
                 g_pAppPlatform->handleKeyEvent(event);
                 break;
+            }
             case SDL_JOYBUTTONDOWN:
             case SDL_JOYBUTTONUP:
+            {
                 // Hate this hack
                 if (event.jbutton.button == SDL_CONTROLLER_BUTTON_START && event.jbutton.state == SDL_PRESSED)
                 {
@@ -176,6 +179,7 @@ static void handle_events()
                 }
                 g_pAppPlatform->handleControllerButtonEvent(event.jbutton.which, event.jbutton.button, event.jbutton.state);
                 break;
+            }
             case SDL_MOUSEBUTTONDOWN:
             case SDL_MOUSEBUTTONUP:
                 if (event.button.button == SDL_BUTTON_WHEELUP)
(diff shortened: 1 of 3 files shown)


platforms/sdl/sdl1/main.cpp
```diff
@@ -130,6 +130,17 @@ static void teardown()
     }
 }
 
+// These macros aren't present in older versions of SDL 1.2.
+// In those versions, event.button.button will either never
+// equal these, or it will and scrolling works the same way,
+// so it's safe to define them ourselves.
+#ifndef SDL_BUTTON_WHEELUP
```
BrentDaMage: can you defined these in `thirdparty/SDL/SDL.h`?

platforms/sdl/sdl1/main.cpp
```diff
@@ -171,15 +182,19 @@ static void handle_events()
...
             case SDL_MOUSEBUTTONUP:
-            {
-                const float scale = g_fPointToPixelScale;
-                MouseButtonType type = UsedAppPlatform::GetMouseButtonType(event.button.button);
-                bool state = UsedAppPlatform::GetMouseButtonState(event);
-                float x = event.button.x * scale;
-                float y = event.button.y * scale;
-                Mouse::feed(type, state, x, y);
+                if (event.button.button == SDL_BUTTON_WHEELUP)
+                    Mouse::feed(MOUSE_BUTTON_SCROLLWHEEL, false, Mouse::getX(), Mouse::getY());
```
BrentDaMage: these two lines do the same shit. you're just gonna only be able to scroll up or down.

BrentDaMage (changes requested):

platforms/sdl/sdl1/main.cpp
Un1q32: Could you reword that? I'm not sure what you mean. Which 2 lines are doing the same thing?

Un1q32: Oh wait I see it now

thirdparty/SDL/SDL.h
```diff
@@ -17,3 +17,15 @@
 #else
     #include <SDL2/SDL.h>
 #endif
+
```
BrentDaMage: sorry, can you put this under line 16?

BrentDaMage (changes requested):

Un1q32:
It's scrolling twice per scroll, gotta fix that rq.

Un1q32:
Ok, fixed

BrentDaMage (approved):

PR #624, opened 2026-06-06
Fixup furnace.png

Un1q32:
The old image was a little messed up, the arrow and fire didn't really fit right with the UI, looking like they were made for a totally different furnace background texture.


Un1q32:
Screenshot (49)

BrentDaMage (approved):

PR #627, opened 2026-06-11
D-Pad Fixes & Additions

Vimd0:
https://github.com/user-attachments/assets/582345bc-b24a-45c5-9ef7-bbb20a00c476

- Fixed moving forward, and swiping to the jump button, the player would stop completely

- Fixed not being able to use the forward left/right buttons

- Added a crouch button (based off 0.12.1 alpha)

- Added a m_swapJumpSneak option

- Added a m_buttonSize option

- Added the pocket progress indicator with meshes (from 0.2.1 alpha)

- Added a 10th slot if on touchscreen


BrentDaMage:
There aren't that many slots on an iPhone 4 screen size, right?

Vimd0:
> There aren't that many slots on an iPhone 4 screen size, right?

Pocket theme has 6 slots. Btw I'm still fixing errors.

source/client/gui/screens/OptionsScreen_Console.cpp
```diff
@@ -77,16 +77,19 @@ ControlsPanelScreen::ControlsPanelScreen(Screen* parent, Minecraft& mc) : PanelS
 {
 	Options& options = *mc.getOptions();
 	int currentIndex = -1;
-	int idxSplit = -1, idxController = -1;
+	int idxSplit = -1, idxController = -1, idxSwapJumpSneak = -1, idxButtonSize = -1;
```
BrentDaMage: Call buttonSize hudSize

Un1q32: Doesn't this make less sense, since it doesn't control the size of other hud elements like the hotbar.  Maybe something like dpadSize or controlsSize.

Vimd0: > Doesn't this make less sense, since it doesn't control the size of other hud elements like the hotbar. Maybe something like dpadSize or controlsSize.

I was just going to say. I'll rename it to dpadSize

BrentDaMage: > Doesn't this make less sense, since it doesn't control the size of other hud elements like the hotbar. Maybe something like dpadSize or controlsSize.

True

source/client/gui/Gui.cpp
```diff
@@ -491,6 +493,51 @@ void Gui::renderMessages(bool bShowAll)
 	}
 }
 
+void Gui::_buildFeedbackMeshes()
```
BrentDaMage, line 495: very solid code. good shit

source/client/gui/Gui.cpp
```diff
@@ -665,29 +714,40 @@ void Gui::renderProgressIndicator(int width, int height)
 		{
 			if (breakProgress > 0.0f)
 			{
-				float xPos = input.m_feedbackX;
-				float yPos = input.m_feedbackY;
+				_buildFeedbackMeshes();
 
-				textures.loadAndBindTexture("gui/feedback_outer.png");
```
BrentDaMage, line 671: could you just comment out the old code? or perhaps put the new feedback circle behind an enhancement macro in GameMods.hpp?

source/client/gui/Gui.cpp
```diff
@@ -781,7 +841,7 @@ int Gui::getNumSlots()
     if (mc.getOptions()->getUiTheme() == UI_POCKET)
 		return 6;
 
-	return 9;
+	return 9 + m_pMinecraft->useTouchscreen();
```
BrentDaMage: Implicit conversion of boolean to int. Can you do a ternary operator instead?
`m_pMinecraft->useTouchscreen() ? 1 : 0`

source/client/gui/Gui.hpp
```diff
@@ -50,6 +50,10 @@ class Gui : public GuiComponent
 private:
 	void renderPumpkin(int, int);
 	void renderVignette(float, int, int);
+	void _buildFeedbackMeshes();
+	mce::Mesh m_feedbackOuter;
```
BrentDaMage: private members should go at the bottom with all of the other member variables.
The vertical ordering should go `private, protected, public`

source/client/player/input/TouchscreenInput_TestFps.cpp
```diff
@@ -297,44 +328,92 @@ void TouchscreenInput_TestFps::tick(Player* pPlayer)
...
+			*m_pSneakExcludeRef = RectangleArea(m_fRightSideX, m_fMiddleY,
+				m_fRightSideX + m_fButtonSize, m_fMiddleY + m_fButtonSize);
 	}
+#endif
+
+	IMoveInput::tick(pPlayer);
 }
 
-static void RenderTouchButton(Tesselator* t, PolygonArea* pArea, int srcX, int srcY)
+static void RenderTouchButton(Tesselator* t, PolygonArea* pArea, int srcX, int srcY, int size = -1, float visualScale = 1.0f, float offsetY = 0.0f)
```
BrentDaMage: I know you didn't do this, but could you remove the `static` declaration from this, and name the function `_renderTouchButton` to match our convention for cpp-file-specific methods?

source/renderer/MatrixStack.cpp
```diff
@@ -182,6 +182,7 @@ void MatrixStack::Ref::release()
 
 Matrix* MatrixStack::Ref::operator->()
 {
+	m_pStack->m_bIsDirty = true;
```
BrentDaMage, line 185: Good catch.

BrentDaMage (changes requested):

source/client/gui/Gui.cpp
Vimd0: created

source/client/gui/Gui.cpp
Vimd0: fixed

source/client/gui/Gui.hpp
Vimd0: fixed

source/client/player/input/TouchscreenInput_TestFps.cpp
Vimd0: fixed

source/client/gui/Gui.hpp
```diff
@@ -109,5 +105,13 @@ class Gui : public GuiComponent
 	bool field_A3C;
 	bool m_bRenderMessages;
     bool m_bRenderHunger;
+
+private:
+#ifdef ENH_NEW_FEEDBACK_INDICATOR
+	void _buildFeedbackMeshes();
```
BrentDaMage: put the method back where it was

source/client/gui/Gui.cpp
```diff
@@ -59,7 +59,9 @@ Gui::Gui(Minecraft* pMinecraft)
 
 	m_pMinecraft = pMinecraft;
 
+#ifdef ENH_NEW_FEEDBACK_INDICATOR
 	m_feedbackMeshesBuilt = false;
```
BrentDaMage: set this regardless

source/client/gui/Gui.cpp
```diff
@@ -493,6 +495,7 @@ void Gui::renderMessages(bool bShowAll)
 	}
 }
 
+#ifdef ENH_NEW_FEEDBACK_INDICATOR
 void Gui::_buildFeedbackMeshes()
 {
```
BrentDaMage, line 496: define this function regardless

BrentDaMage (changes requested):

source/client/gui/Gui.cpp
Vimd0: fixed

source/client/gui/Gui.cpp
Vimd0: fixed

source/client/gui/Gui.hpp
Vimd0: fixed

source/client/gui/Gui.hpp
```diff
@@ -50,6 +50,10 @@ class Gui : public GuiComponent
 private:
 	void renderPumpkin(int, int);
 	void renderVignette(float, int, int);
+	void _buildFeedbackMeshes();
+	bool m_feedbackMeshesBuilt;
```
BrentDaMage: these fields should go with the other private fields

BrentDaMage (changes requested):

source/client/gui/Gui.hpp
```diff
@@ -42,6 +42,9 @@ class Gui : public GuiComponent
 
 private:
 	static bool _isVignetteAvailable;
+	bool m_feedbackMeshesBuilt;
```
BrentDaMage: these are not static, these do not do here. put them down with the rest of the member variables at the bottom of the file, in order of accessibility, as described in one of the previous comments.

BrentDaMage (changes requested):

source/client/gui/Gui.hpp
```diff
@@ -109,5 +106,8 @@ class Gui : public GuiComponent
 	bool field_A3C;
 	bool m_bRenderMessages;
     bool m_bRenderHunger;
+	bool m_feedbackMeshesBuilt;
```
BrentDaMage: make this private

BrentDaMage (changes requested):

source/client/gui/Gui.hpp
Vimd0: fixed

PR #645, opened 2026-06-27
BucketItem

### LINKED ISSUE #622: Creative inventory adds to held stacks of the same item id with different damage values
If you hold an oak log in the creative inventory and click on the creative inventories' birch log, it should destroy the oak log like it does when clicking on other unrelated items, but instead it gives you a second oak log.  The same thing happens with dyes.

Wilyicaro:
- Implements `BucketItem`, as the title says
- Refactors `Item::use` and `ItemStack::use` to be more consistent

### DIFF
--- a/source/world/entity/Cow.cpp
+++ b/source/world/entity/Cow.cpp
@@ -6,6 +6,7 @@
 	SPDX-License-Identifier: BSD-1-Clause
  ********************************************************************/
 #include "Cow.hpp"
+#include "Player.hpp"
 
 Cow::Cow(Level* pLevel) : Animal(pLevel)
 {
@@ -23,4 +24,16 @@ void Cow::addAdditionalSaveData(CompoundTag& tag) const
 void Cow::readAdditionalSaveData(const CompoundTag& tag)
 {
 	Animal::readAdditionalSaveData(tag);
+}
+
+bool Cow::interact(Player* player)
+{
+	ItemStack& selected = player->m_pInventory->getSelected();
+	if (selected.getId() == Item::bucket_empty->m_itemID)
+	{
+		player->m_pInventory->setSelectedItem(Item::milk);
+		return true;
+	}
+	else
+		return false;
 }
--- a/source/world/entity/Cow.cpp
+++ b/source/world/entity/Cow.cpp
@@ -34,6 +34,6 @@ bool Cow::interact(Player* player)
 		player->m_pInventory->setSelectedItem(Item::milk);
 		return true;
 	}
-	else
-		return false;
+		
+	return false;
 }
--- a/source/world/item/BowItem.cpp
+++ b/source/world/item/BowItem.cpp
@@ -8,13 +8,13 @@ BowItem::BowItem(int id) : Item(id)
 	m_maxStackSize = 1;
 }
 
-bool BowItem::use(ItemStack& item, Level* level, Player* user) const
+bool BowItem::use(ItemStack& item, Level* level, Mob& mob) const
 {
-    if (!user->isPlayer() || user->isCreative() || user->m_pInventory->removeResource(Item::arrow->m_itemID))
+    if (!mob.isPlayer() || static_cast<Player&>(mob).isCreative() || static_cast<Player&>(mob).m_pInventory->removeResource(Item::arrow->m_itemID))
     {
-        level->playSound(user, "random.bow", 1.0f, 1.0f / (level->m_random.nextFloat() * 0.4f + 0.8f));
+        level->playSound(&mob, "random.bow", 1.0f, 1.0f / (level->m_random.nextFloat() * 0.4f + 0.8f));
         if (!level->m_bIsClientSide)
-            level->addEntity(new Arrow(level, user));
+            level->addEntity(new Arrow(level, &mob));
     }
 
     return false;
--- a/source/world/item/BowItem.cpp
+++ b/source/world/item/BowItem.cpp
@@ -8,14 +8,14 @@ BowItem::BowItem(int id) : Item(id)
 	m_maxStackSize = 1;
 }
 
-ItemStack* BowItem::use(ItemStack* inst, Level* level, Mob* user) const
+bool BowItem::use(ItemStack& item, Level* level, Player* user) const
 {
-    if (!user->isPlayer() || static_cast<Player*>(user)->isCreative() || static_cast<Player*>(user)->m_pInventory->removeResource(Item::arrow->m_itemID))
+    if (!user->isPlayer() || user->isCreative() || user->m_pInventory->removeResource(Item::arrow->m_itemID))
     {
         level->playSound(user, "random.bow", 1.0f, 1.0f / (level->m_random.nextFloat() * 0.4f + 0.8f));
         if (!level->m_bIsClientSide)
             level->addEntity(new Arrow(level, user));
     }
 
-    return inst;
+    return false;
 }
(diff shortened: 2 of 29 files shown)


source/world/entity/Cow.cpp
```diff
@@ -24,3 +25,15 @@ void Cow::readAdditionalSaveData(const CompoundTag& tag)
...
+
+bool Cow::interact(Player* player)
+{
+	ItemStack& selected = player->m_pInventory->getSelected();
+	if (selected.getId() == Item::bucket_empty->m_itemID)
+	{
+		player->m_pInventory->setSelectedItem(Item::milk);
+		return true;
+	}
+	else
```
BrentDaMage: remove else, just return false

source/world/item/BowItem.cpp
```diff
@@ -8,14 +8,14 @@ BowItem::BowItem(int id) : Item(id)
...
+    if (!user->isPlayer() || user->isCreative() || user->m_pInventory->removeResource(Item::arrow->m_itemID))
     {
         level->playSound(user, "random.bow", 1.0f, 1.0f / (level->m_random.nextFloat() * 0.4f + 0.8f));
         if (!level->m_bIsClientSide)
             level->addEntity(new Arrow(level, user));
     }
 
-    return inst;
+    return false;
 }
```
BrentDaMage: new line at bottom of file

source/world/item/BucketItem.cpp
```diff
@@ -0,0 +1,73 @@
...
+{
+	m_maxStackSize = 1;
+    m_maxDamage = 64;
+    m_content = content;
+}
+
+bool BucketItem::use(ItemStack& item, Level* level, Player* player) const
+{
+    Vec3 headPos(player->m_pos.x, player->m_pos.y + 1.62 - player->m_heightOffset, player->m_pos.z);
+    float constexpr r = M_PI / 180;
```
BrentDaMage: put constexpr before float

source/world/item/BucketItem.cpp
```diff
@@ -0,0 +1,73 @@
...
+    {
+        if (hitResult.m_hitType == HitResult::TILE)
+        {
+            TilePos tp(hitResult.m_tilePos);
+            if (!level->mayInteract(player, tp))
+                return false;
+
+            if (m_content == TILE_AIR)
+            {
+                if (level->getMaterial(tp) == Material::water && level->getData(tp) == 0)
```
BrentDaMage: store the `level->getMaterial()` and `level->getData()` calls into variables before checking.
material should be stored above line 33

source/world/item/BucketItem.cpp
```diff
@@ -0,0 +1,73 @@
...
+BucketItem::BucketItem(int id, TileID content) : Item(id)
+{
+	m_maxStackSize = 1;
+    m_maxDamage = 64;
+    m_content = content;
+}
+
+bool BucketItem::use(ItemStack& item, Level* level, Player* player) const
+{
+    Vec3 headPos(player->m_pos.x, player->m_pos.y + 1.62 - player->m_heightOffset, player->m_pos.z);
```
BrentDaMage: don't forget the f

source/world/item/EggItem.cpp
```diff
@@ -9,11 +9,11 @@ EggItem::EggItem(int itemID) : Item(itemID)
 	m_maxStackSize = 16;
 }
 
-ItemStack* EggItem::use(ItemStack* inst, Level* level, Mob* user) const 
+bool EggItem::use(ItemStack& inst, Level* level, Player* user) const 
 {
-	--inst->m_count;
+	--inst.m_count;
 	level->playSound(user, "random.bow", 0.5f, 0.4F / (level->m_random.nextFloat() * 0.4F + 0.8F));
```
BrentDaMage: make these fs lowercase please

source/world/item/FoodItem.cpp
```diff
@@ -6,12 +6,13 @@ FoodItem::FoodItem(int id, int nutrition) : Item(id), m_nutrition(nutrition)
...
     {
-        --inst->m_count;
-        mob->heal(m_nutrition);
+        --inst.m_count;
+        player->heal(m_nutrition);
+        return true;
     }
-    return inst;
+    return false;
 }
```
BrentDaMage: add new line at end of file

source/world/item/Item.cpp
```diff
@@ -618,9 +620,9 @@ float Item::getDestroySpeed(ItemStack* instance, const Tile* tile) const
 	return 1.0f;
 }
 
-ItemStack* Item::use(ItemStack* instance, Level* level, Mob* user) const
+bool Item::use(ItemStack& item, Level* level, Player* user) const
```
BrentDaMage: Can we keep this as a Mob, or better yet, make it a Mob& instead? We already have the code written for checking if the user is a Player or not, I don't see a reason to change it now. Plus, how cool would it be to have mobs be able to just use any item?

source/world/item/SnowballItem.hpp
```diff
@@ -8,5 +8,5 @@ class SnowballItem : public Item
 	SnowballItem(int);
 
 public:
-	ItemStack* use(ItemStack* inst, Level* level, Mob* user) const;
+	bool use(ItemStack& inst, Level* level, Player* user) const override;
```
BrentDaMage: good catch

BrentDaMage (changes requested):

source/world/item/BucketItem.cpp
```diff
@@ -0,0 +1,79 @@
...
+	m_maxStackSize = 1;
+    m_maxDamage = 64;
+    m_content = content;
+}
+
+bool BucketItem::use(ItemStack& item, Level* level, Mob& mob) const
+{
+    if (!mob.isPlayer()) return false;
+
+    Player* player = static_cast<Player*>(&mob);
```
BrentDaMage: why change to a pointer?

source/world/item/Item.hpp
```diff
@@ -55,7 +55,7 @@ class Item
 	virtual bool useOn(ItemStack*, Level*, const TilePos& pos, Facing::Name face) const;
 	virtual bool useOn(ItemStack*, Player*, Level*, const TilePos& pos, Facing::Name face) const;
 	virtual float getDestroySpeed(ItemStack*, const Tile*) const;
-	virtual ItemStack* use(ItemStack*, Level*, Mob*) const;
+	virtual bool use(ItemStack&, Level*, Mob& mob) const;
```
BrentDaMage: Can we keep calling the Mob& parameter `user`? I feel like it makes the Mob's role more clear than `mob`.

BrentDaMage (changes requested):

source/world/item/BucketItem.cpp
Wilyicaro: makes the change easier, I would have to pass it as a pointer anyway in some of the called functions

BrentDaMage: fair enough

source/world/item/Item.hpp
Wilyicaro: Generally, `user` is associated with a player, even though, grammatically, it's not necessarily so.

BrentDaMage: Right, but I think it might be unclear if Mob is the thing using the item, or if it's what the item is being used *on*.

BrentDaMage (approved):

PR #646, opened 2026-06-27
Nix Flake support for Nix users

MolassesLover:
This pull request adds a Nix Flake, providing the `nbcraft` package for easy installation through the Nix package manager, Home Manager, and NixOS. This package consequently also provides reproducible builds, as such Nix users can also test, debug, and inspect NBcraft in a common environment. I believe this change improves the availability of NBcraft a lot.

This pull request also changes the CMake build process, adding an installation target.
The `stb` source and include paths are also now variables so that they can be pointed to Nix's `stb` files during the build process.

```cmake
set(STB_IMAGE_SOURCE src/stb_image_impl.c)

if(NOT NIX_BUILD)
    set(STB_VORBIS_SOURCE include/stb_vorbis.c)
    set(STB_INCLUDE include)
endif()
```

Lastly, some `include`s have been change in order to support system-wide `stb` installations, such as through Nix. This does not conflict with the existing submodule-based approach, which remains the default. Nix Flakes unfortunately don't properly support Git submodules, making this change necessary.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -143,3 +143,6 @@ xcuserdata/
 
 # Clangd
 .clangd
+
+# Nix derivation result outputs
+[Rr]esult
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,6 +5,19 @@ project(nbcraft)
 set(CMAKE_LIBRARY_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}")
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}")
 
+option(NBC_SYSTEM_STB OFF)
+option(NBC_NIX_BUILD OFF)
+
+if(NBC_NIX_BUILD)
+    if(NOT NBC_SYSTEM_STB)
+        message(SEND_ERROR "Option NBC_NIX_BUILD depends on NBC_SYSTEM_STB, needs to be ON.")
+    endif()
+endif()
+
+if(NBC_SYSTEM_STB)
+    add_compile_definitions(NBC_SYSTEM_STB)
+endif()
+
 # MC_VERSION string is an 7 digit number with 4 components, era, major, minor, and patch.
 # Each component takes 2 digits, except era, which takes 1. Major, minor, and patch are self explanatory,
 # era designates if the version is from alpha, beta, or release. Alpha is 1, beta is 2, and release is 3.
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,35 @@
+{
+  description = "The ultimate cross-platform mining experience.";
+
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+  };
+
+  outputs =
+    { self, nixpkgs }:
+    let
+      supportedSystems = [
+        "aarch64-linux"
+        "i686-linux"
+        "x86_64-linux"
+        "aarch64-darwin"
+        "x86_64-darwin"
+      ];
+
+      forAllSystems = nixpkgs.lib.genAttrs supportedSystems;
+    in
+    {
+      packages = forAllSystems (
+        system:
+        let
+          pkgs = (import nixpkgs { inherit system; });
+          nbcraft = pkgs.callPackage platforms/nix/package.nix { };
+        in
+        {
+          inherit nbcraft;
+          default = nbcraft;
+          devShell = pkgs.mkShell { inputsFrom = [ nbcraft ]; };
+        }
+      );
+    };
+}
--- a/platforms/nix/package.nix
+++ b/platforms/nix/package.nix
@@ -1,0 +1,55 @@
+{
+  lib,
+  stdenv,
+  cmake,
+  ninja,
+  fetchFromGitHub,
+  xorg,
+  xorgserver,
+  pkg-config,
+  SDL2,
+  openal,
+  libGLX,
+  libvorbis,
+  stb,
+  clang,
+}:
+
+stdenv.mkDerivation (finalAttrs: {
+  pname = "nbcraft";
+  version = "0.3.3";
+
+  src = ../../.;
+
+  nativeBuildInputs = [
+    cmake
+    ninja
+    clang
+  ];
+  buildInputs = [
+    SDL2
+    openal
+    libGLX
+    libvorbis
+    # Build a newer version of stb
+    (stb.overrideAttrs (previousAttrs: {
+      version = "0-unstable-2026-04-15";
+      src = fetchFromGitHub {
+        owner = "nothings";
+        repo = "stb";
+        rev = "31c1ad37456438565541f4919958214b6e762fb4";
+        hash = "sha256-m2yNUlA37hDkKQVrQ+R8nufHfW/cXLnMo+n1X1Cyun0=";
+      };
+    }))
+  ];
+
+  cmakeFlags = [
+    "-D NBC_SYSTEM_STB=ON" # Required by NBC_NIX_BUILD
+    "-D NBC_NIX_BUILD=ON"
+    "-D CMAKE_C_COMPILER=${clang}/bin/clang"
+    "-D CMAKE_CXX_COMPILER=${clang}/bin/clang++"
+    "-D CMAKE_CXX_STANDARD_INCLUDE_DIRECTORIES=${stb}/include/stb/"
+    "-D STB_INCLUDE=${stb}/include/stb/"
+    "-D STB_VORBIS_SOURCE=${stb}/include/stb/stb_vorbis.c"
+  ];
+})
--- a/platforms/sdl/base/AppPlatform_sdl.cpp
+++ b/platforms/sdl/base/AppPlatform_sdl.cpp
@@ -10,8 +10,13 @@
 #include <emscripten.h>
 #endif
 
-#include "thirdparty/stb_image/include/stb_image.h"
-#include "thirdparty/stb_image/include/stb_image_write.h"
+#if NBC_SYSTEM_STB
+    #include <stb_image.h>
+	#include <stb_image_write.h>
+#else
+	#include "thirdparty/stb_image/include/stb_image.h"
+	#include "thirdparty/stb_image/include/stb_image_write.h"
+#endif
 
 #if MCE_GFX_API_OGL
 // needed for screenshots
--- a/platforms/windows/CMakeLists.txt
+++ b/platforms/windows/CMakeLists.txt
@@ -46,3 +46,5 @@ target_link_libraries(nbcraft nbcraft-core)
 
 # stb_image (If Needed)
 target_link_libraries(nbcraft stb_image)
+
+install(TARGETS nbcraft DESTINATION bin)
--- a/thirdparty/glm/CMakeLists.txt
+++ b/thirdparty/glm/CMakeLists.txt
@@ -36,6 +36,8 @@ if(GLM_TEST_ENABLE)
 		${CORE_SOURCE}    ${CORE_INLINE}    ${CORE_HEADER}
 		${GTC_SOURCE}     ${GTC_INLINE}     ${GTC_HEADER}
 		${GTX_SOURCE}     ${GTX_INLINE}     ${GTX_HEADER})
+
+	install(TARGETS ${NAME} DESTINATION bin)
 endif(GLM_TEST_ENABLE)
 
 #add_library(glm STATIC glm.cpp)
--- a/thirdparty/stb_image/CMakeLists.txt
+++ b/thirdparty/stb_image/CMakeLists.txt
@@ -1,8 +1,15 @@
 cmake_minimum_required(VERSION 3.16.0)
 project(stb_image)
 
+set(STB_IMAGE_SOURCE src/stb_image_impl.c)
+
+if(NOT NBC_NIX_BUILD)
+    set(STB_VORBIS_SOURCE include/stb_vorbis.c)
+    set(STB_INCLUDE include)
+endif()
+
 # Build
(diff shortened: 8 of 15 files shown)


bluepilledgreat:
Visual Studio and X-Code projects in https://github.com/nbcraft-org/nbcraft/tree/master/projects need to be updated

Un1q32:
> Visual Studio and X-Code projects in https://github.com/nbcraft-org/nbcraft/tree/master/projects need to be updated

If they were already including the one thing it should still work, the makefile build still works without modifications.

bluepilledgreat:
> If they were already including the one thing it should still work, the makefile build still works without modifications.

They are not directly including the STB library

misc/nix/package.nix
```diff
@@ -0,0 +1,34 @@
+{ lib, stdenv, cmake, ninja, fetchFromGitHub, xorg, xorgserver, pkg-config
+, SDL2, openal, libGLX, libvorbis, stb, clang }:
+
+stdenv.mkDerivation (finalAttrs: {
+  pname = "nbcraft";
+  version = "0.1.0";
```
BrentDaMage: This should match the version in the code, which currently is 0.3.3.

misc/nix/package.nix
```diff
@@ -0,0 +1,34 @@
...
+  ];
+
+  cmakeFlags = [
+    "-D NIX_BUILD=ON"
+    "-D CMAKE_C_COMPILER=${clang}/bin/clang"
+    "-D CMAKE_CXX_COMPILER=${clang}/bin/clang++"
+    "-D CMAKE_CXX_STANDARD_INCLUDE_DIRECTORIES=${stb}/include/stb/"
+    "-D STB_INCLUDE=${stb}/include/stb/"
+    "-D STB_VORBIS_SOURCE=${stb}/include/stb/stb_vorbis.c"
+    "-D CMAKE_BUILD_TYPE=Debug"
```
BrentDaMage: Correct me if I'm wrong, but wouldn't this just always end up producing a debug build?

platforms/nix/package.nix
BrentDaMage, line 1: Please put the `nix` folder in `platforms/linux/`

platforms/windows/CMakeLists.txt
```diff
@@ -46,3 +46,5 @@ target_link_libraries(nbcraft nbcraft-core)
 
 # stb_image (If Needed)
 target_link_libraries(nbcraft stb_image)
+
+install(TARGETS nbcraft DESTINATION bin)
```
BrentDaMage: add new line at bottom of file to satisfy the git gods

thirdparty/glm/CMakeLists.txt
```diff
@@ -36,7 +36,9 @@ if(GLM_TEST_ENABLE)
 		${CORE_SOURCE}    ${CORE_INLINE}    ${CORE_HEADER}
 		${GTC_SOURCE}     ${GTC_INLINE}     ${GTC_HEADER}
 		${GTX_SOURCE}     ${GTX_INLINE}     ${GTX_HEADER})
+
+	install(TARGETS ${NAME} DESTINATION bin)
 endif(GLM_TEST_ENABLE)
 
 #add_library(glm STATIC glm.cpp)
-#add_library(glm_shared SHARED glm.cpp)
+#add_library(glm_shared SHARED glm.cpp)
```
BrentDaMage: add new line

CMakeLists.txt
```diff
@@ -5,6 +5,8 @@ project(nbcraft)
 set(CMAKE_LIBRARY_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}")
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}")
 
+option(NIX_BUILD OFF)
```
BrentDaMage: call this `NBC_NIX_BUILD`

thirdparty/stb_image/CMakeLists.txt
```diff
@@ -1,8 +1,15 @@
...
+
 # Build
-add_library(stb_image STATIC src/stb_image_impl.c)
-target_include_directories(stb_image PUBLIC include)
-add_library(stb_vorbis STATIC include/stb_vorbis.c)
-target_include_directories(stb_vorbis PUBLIC include)
+add_library(stb_image STATIC ${STB_IMAGE_SOURCE})
+target_include_directories(stb_image PUBLIC ${STB_INCLUDE})
+add_library(stb_vorbis STATIC ${STB_VORBIS_SOURCE})
+target_include_directories(stb_vorbis PUBLIC ${STB_INCLUDE})
```
BrentDaMage: add new line at bottom of file

.gitignore
```diff
@@ -10,6 +10,7 @@
 [Dd]ebugPublic/
 [Rr]elease/
 [Rr]eleases/
+[Rr]esult
```
BrentDaMage: if this is not a part of an official GitHub gitignore, please add it at the bottom of the file and create a comment indicating what it's for.

flake.nix
```diff
@@ -0,0 +1,23 @@
+{
+  description =
+    "An unofficial, community run, and open-source re-implementation of Minecraft: Pocket Edition (v0.1.3). Unaffiliated with Mojang.";
```
BrentDaMage: I understand this description is more accurate to what we're doing here, but in the interest of avoiding angering Mojang, could we have it use the description from the Flatpak config file/GitHub repo?
`The ultimate cross-platform mining experience.`

flake.nix
BrentDaMage, line 1: Would it be possible to move this and the .lock file to `platforms/linux/nix`?

BrentDaMage (changes requested):

platforms/nix/package.nix
Un1q32: Doesn't the nix package manager support macOS?  Something like platforms/nix might be better.

flake.nix
MolassesLover: This would complicate installations a lot when using Nix configuration files in something like Home Manager, NixOS, custom Flakes, etc. It's generally best practice to keep this in the root of the source tree, sorry for the inconvenience.

BrentDaMage: Alright then.

MolassesLover:
Apologies for the messy pull request. Hope that commit [c7afa93](https://github.com/nbcraft-org/nbcraft/pull/646/commits/c7afa93c0d3c0e4337e9158c20bd1563ae76b64b) addresses the majority of issues pointed out.

MolassesLover:
> > If they were already including the one thing it should still work, the makefile build still works without modifications.
> 
> They are not directly including the STB library

I'm not entirely sure I understand, could you clarify what the issue is here? Are there `include` paths that I need to alter? I'm not that familiar with Visual Studio and X-Code solutions, as I generally work on Linux, and prefer to use CMake from a CLI across other OSes.

MolassesLover:
> I'm not entirely sure I understand, could you clarify what the issue is here? Are there `include` paths that I need to alter? I'm not that familiar with Visual Studio and X-Code solutions, as I generally work on Linux, and prefer to use CMake from a CLI across other OSes.

I'm under the impression these project files use the Git submodule paths? They should still build, I think?

flake.nix
```diff
@@ -1,23 +1,32 @@
...
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+  };
 
-  outputs = { self, nixpkgs }:
+  outputs =
+    { self, nixpkgs }:
     let
-      supportedSystems = [ "x86_64-linux" "aarch64-linux" ];
+      supportedSystems = [
```
BrentDaMage, line 11: what about 32-bit x86 and PowerPC?

platforms/nix/package.nix
BrentDaMage: Nevermind, looks like it supports more than just Linux.

BrentDaMage (changes requested):
Fix build errors

flake.nix
Fierelier: Brent ur a boss

MolassesLover: No official PowerPC support, unfortunately. Commit [`e7b34b`](https://github.com/MolassesLover/nbcraft-molasses/commit/e7b34bb2f48304b301485068ed15f6c77d621fa2) enables support for all official Nix systems.

BrentDaMage:
Do not rebase, merge instead.

bluepilledgreat:
295 changed files for Nix support is crazy work

MolassesLover:
Waiting for the workflows to run before I make any more changes, but is there any interest in having workflows for the Nix builds?

MolassesLover:
> Do not rebase, merge instead.

Can't you just squash commit? That seems more reasonable than having me force push.

MolassesLover:
> > Do not rebase, merge instead.
> 
> Can't you just squash commit? That seems more reasonable than having me force push.

Git doesn't really mind this sort of thing. Force pushing is more problematic, given it allows deletions.

Merging `nbcraft-org:master` with `MolassesLover:nix-packaging` by default looks like this:
```
Auto-merging source/client/sound/SoundData.cpp
Merge made by the 'ort' strategy.
 .gitignore                              |  3 +++
 CMakeLists.txt                          | 13 +++++++++++++
 flake.lock                              | 27 +++++++++++++++++++++++++++
 flake.nix                               | 35 +++++++++++++++++++++++++++++++++++
 platforms/nix/package.nix               | 55 +++++++++++++++++++++++++++++++++++++++++++++++++++++++
 platforms/sdl/base/AppPlatform_sdl.cpp  |  9 +++++++--
 platforms/sdl/sdl1/CMakeLists.txt       |  4 +++-
 platforms/sdl/sdl2/CMakeLists.txt       |  1 +
 platforms/windows/AppPlatform_win32.cpp |  9 +++++++--
 platforms/windows/CMakeLists.txt        |  2 ++
 source/client/app/AppPlatform.cpp       |  6 +++++-
 source/client/sound/SoundData.cpp       |  7 ++++++-
 source/client/sound/SoundStream.hpp     |  7 ++++++-
 thirdparty/glm/CMakeLists.txt           |  2 ++
 thirdparty/stb_image/CMakeLists.txt     | 15 +++++++++++----
 15 files changed, 183 insertions(+), 12 deletions(-)
 create mode 100644 flake.lock
 create mode 100644 flake.nix
 create mode 100644 platforms/nix/package.nix
```

BrentDaMage:
It says you changed 290-something files. I don't think squashing will magically get rid of that. Usually git is good enough to where it will recognize the fact that nothing's changed, but the fact it thinks 295 files have changed indicates an issue.

MolassesLover:
> It says you changed 290-something files. I don't think squashing will magically get rid of that. Usually git is good enough to where it will recognize the fact that nothing's changed, but the fact it thinks 295 files have changed indicates an issue.

Git thinks 295 files have changed because across *my repository's commit history 295 files have changed*. GitHub here is using my repository's commit history to determine what files have changed, and it's right to say that these files have actually changed.

Regardless, merging with default settings will ignore these changed files, given the changes are already part of this remote's commit history. As you can see in the Git output I posted, it understands that it doesn't have to merge commits which don't actually affect files that are identical.

My pull request's [changed files tab](https://github.com/nbcraft-org/nbcraft/pull/646/changes) doesn't depend on the amount of files which might change when merging, given it doesn't know how it is that you want to merge. 

If you're scared of merging with the default settings, you can manually squash commit. In the case of a squash commit, Git would ignore my repository's commit history, and instead boil it down to an entire commit. That way Git doesn't have to worry about merging every commit.

BrentDaMage:
Could you put the PR in a state where I can reasonably review it?
image

MolassesLover:
> Could you put the PR in a state where I can reasonably review it?

Alright, force pushed and deleted the rebase commits along with their objects. I hope this is easier to review from GitHub.

If necessary, you can find the original commit history [here](https://github.com/MolassesLover/nbcraft-molasses/commits/nix-packaging-old).

platforms/sdl/base/AppPlatform_sdl.cpp
```diff
@@ -10,8 +10,13 @@
 #include <emscripten.h>
 #endif
 
-#include "thirdparty/stb_image/include/stb_image.h"
-#include "thirdparty/stb_image/include/stb_image_write.h"
+#if NBC_SYSTEM_STB
+    #include <stb_image.h>
+	#include <stb_image_write.h>
+#else
+	#include "thirdparty/stb_image/include/stb_image.h"
```
BrentDaMage, line 17: Rename the `stb_image` directory (that contains the Git submodule) to `stb`.
Then, create the following files with the `NBC_SYSTEM_STB` check included, as listed.
```
thirdparty/stb_image/include/stb_image.h
thirdparty/stb_image/include/stb_image_write.h
thirdparty/stb_image/include/stb_vorbis.c
```
This will remove the need to check this macro every time we want to include an STB header, as well as remove the need to change any pre-existing STB header includes.

CMakeLists.txt
```diff
@@ -5,6 +5,19 @@ project(nbcraft)
 set(CMAKE_LIBRARY_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}")
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}")
 
+option(NBC_SYSTEM_STB OFF)
+option(NBC_NIX_BUILD OFF)
+
+if(NBC_NIX_BUILD)
+    if(NOT NBC_SYSTEM_STB)
+        message(SEND_ERROR "Option NBC_NIX_BUILD depends on NBC_SYSTEM_STB, needs to be ON.")
```
BrentDaMage, line 13: Why not just force this on automatically?

BrentDaMage (changes requested):

PR #651, opened 2026-06-29
Add mssing textures for nether mobs and fix some warnings

Un1q32:
The warnings weren't caught before by CI because they only appear in newer C++ modes I think.

The textures, especially the ghast one, are a bit odd, but its better than just showing a missing texture.  Maybe someone will draw new textures later.


BrentDaMage (approved):
textures kinda look like doom

PR #663, opened 2026-07-02
Add file utility class

bluepilledgreat:


bluepilledgreat:
i don't have xcode so can someone add the files to that project

PR #665, opened 2026-07-03
Add multiple absent console UI screens, features and other misc fixes and tweaks.

fileshredder:
This PR aims to add multiple screens that were missing their X360 counterparts in the project, and instead were using the normal Java/PE screens.

**Console UI additions:**

- Now Exiting the world will show the Console Edition Saving World Screen instead of the normal one.
- Starting the game up in Console UI mode will now display the autosave warning screen, which will close in 10 seconds if not given a response
- Now selecting a player in the Online Worlds part of the Select World List will open a replica of the Console Join Game screen instead of directly connecting to the player.
- Selecting a world in the world list now opens the Save Select Menu where you can see the gamemode the world was created in, it's seed and adjust the game's difficulty or load the world.
- Now the player's nametag is outlined when playing multiplayer like in console edition (hardcoded to green currently.)
- Items (not blocks) now no longer have normal lighting/shading when held, matching their look that was in TU0/1/2.

**Console UI changes:**

- Hotbar position has been adjusted to match TU2 as closely as possible
- Chat Messages have been accurately adjusted to match TU2's messages, no longer having text shadows and their layout position matching TU2's
- The F3 Debug Screen no longer renders abnormally small in Console UI mode and will now render at the correct normal size instead.
- The translucent background in screens like the pause screen will now in Console UI mode, have less opacity like TU2
~~- Font Spacing changes in Console UI mode to match how it was in TU2~~
- Button text positioning and shadow distance has been adjusted to match TU2
- Select World Screen now uses the correct panel type instead of SMALL_PANEL
- Added PANEL_SQUARE_RECCESS
- Crafting Screen layout slightly adjusted to be almost pixel accurate to TU2
- Positioning of blocks in hotbar and inventory have been slightly adjusted
- Pause screen layout has been fixed to match TU2.

**Misc Fixes:**

- Fixed Fire Tesselation code so that it now renders correctly on some sides and is no longer invisible instead.
- Flint and Steel no longer stacks to 16 and now has durability.
- External File Level Storage now also has the ability to read the seed of a world.

image
image
image
image
image
image
image
image


game/assets/lang/en_us.json
```diff
@@ -146,5 +146,15 @@
...
+    "container.tab.armour": "Armour",
+    "loadMenu.online": "Online game",
+    "loadMenu.inviteOnly": "Invite only",
+    "loadMenu.load": "Load",
+    "loadMenu.gamemodeInfo": "Created in %s Mode",
+    "loadMenu.seed": "Seed:  %s",
+    "joinMenu.join": "Join Game",
+    "saving.save": "Preparing to Save Level",
+    "saving.progress": "Preparing Chunks...",
+    "autosaveWarning.warning": "This game has a level autosave feature. When you see the icon above displayed, the game is saving your data.\n Please do not turn off your Xbox 360 console while this icon is on-screen.",
```
BrentDaMage: Change "Xbox 360 console" to "device".

source/client/gui/components/Button.cpp
```diff
@@ -81,9 +82,10 @@ void Button::_renderMessageConsole(Font& font)
 
 	if (!isEnabled())
 		textColor.a *= 0.5f;
+//		textColor = Color(167, 167, 167); // PARITY-XBOX: In TU2 the text is colored gray instead of having reduced transparency.
```
BrentDaMage: Do `@PARITY-LCE` instead

source/client/gui/screens/AutosaveWarningScreen_Console.cpp
```diff
@@ -0,0 +1,82 @@
+#include "AutosaveWarningScreen_Console.hpp"
+#include "common/Util.hpp"
+#include "client/locale/Language.hpp"
+#include "CreateWorldScreen.hpp"
+
+AutosaveWarningScreen_Console::AutosaveWarningScreen_Console(Screen* parent) :
```
BrentDaMage: : and , should be on the start of each new line, before each member variable

source/client/gui/screens/AutosaveWarningScreen_Console.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+	m_autosaveLine = m_pFont->split(text, maxTextWidth);
+}
+
+void AutosaveWarningScreen_Console::render(float f)
+{
+	PanelScreen_Console::render(f);
+	Font& font = *m_pFont;
+
+	int yOffs = std::sin(getTimeMs() / 210.0f) * 4.0f;
```
BrentDaMage: Can you use the `m_timer` in `m_pMinecraft` instead? Using that would be more efficient, since with getTimeMs, we are querying the current time from the CPU down to the microsecond (at least on UNIX).

source/client/gui/screens/AutosaveWarningScreen_Console.cpp
```diff
@@ -0,0 +1,82 @@
...
+}
+
+void AutosaveWarningScreen_Console::tick()
+{
+	if (m_timer < 0)
+		return;
+
+	m_timer++;
+
+	if (m_timer == 200)
```
BrentDaMage: Make `200` a macro at the top of the file, prefixed with `C_`

source/client/gui/screens/AutosaveWarningScreen_Console.hpp
```diff
@@ -0,0 +1,29 @@
...
+protected:
+	void _buttonClicked(Button* pButton) override;
+
+public:
+	void init() override;
+	void render(float f) override;
+	void renderPanel(float f) override;
+	void tick() override;
+
+public:
```
BrentDaMage, line 25: Private members should come before public members. It should be in order from least accessible to most accessible.

source/client/gui/screens/AutosaveWarningScreen_Console.hpp
```diff
@@ -0,0 +1,29 @@
+#pragma once
+
+#include "PanelScreen_Console.hpp"
+#include "client/gui/components/Button.hpp"
+#include "client/gui/components/TickBox.hpp"
```
BrentDaMage: Only include these headers if needed. If they are needed, include them in the cpp file instead.

source/client/gui/screens/JoinGameScreen_Console.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+	currentShaderColor = Color::WHITE;
+}
+
+void UsernameButton::renderMessage(Font& font, const Color& textColor)
+{
+	font.drawScalableShadow(getMessage(), m_xPos + 10, m_yPos + (m_height - 16) / 2, textColor);
+}
+
+JoinGameScreen_Console::JoinGameScreen_Console(Screen* parent, const PingedCompatibleServer& server) :
```
BrentDaMage:
```cpp
JoinGameScreen_Console::JoinGameScreen_Console(Screen* parent, const PingedCompatibleServer& server)
	: PanelScreen_Console(parent)
	, m_btnJoinGame(0, 0, 450, 40, Language::get("joinMenu.join"))
	, m_btnUsername(server.m_name.C_String())
	, m_server(server)
	, m_panelAlpha(1.0f)
```

source/client/gui/screens/JoinGameScreen_Console.cpp
```diff
@@ -0,0 +1,98 @@
...
+void JoinGameScreen_Console::render(float f)
+{
+	PanelScreen_Console::render(f);
+}
+
+void JoinGameScreen_Console::renderPanel(float f)
+{
+	float targetAlpha = m_btnJoinGame.isSelected() ? 0.5f : 1.0f;
+
+	float fadeSpeed = 0.05f;
```
BrentDaMage: make this a constexpr

source/client/gui/screens/JoinGameScreen_Console.hpp
```diff
@@ -0,0 +1,37 @@
+#pragma once
+
+#include "PanelScreen_Console.hpp"
+#include "client/gui/components/Button.hpp"
+#include "client/gui/components/TickBox.hpp"
+#include "client/gui/components/SliderButton.hpp"
+#include "client/gui/components/TextBox.hpp"
+class UsernameButton : public Button
```
BrentDaMage, line 6: put new line above class def

source/client/gui/screens/JoinGameScreen_Console.hpp
```diff
@@ -0,0 +1,37 @@
+#pragma once
+
+#include "PanelScreen_Console.hpp"
+#include "client/gui/components/Button.hpp"
+#include "client/gui/components/TickBox.hpp"
```
BrentDaMage: Include these in the cpp file, don't include them if not needed.

source/client/gui/screens/JoinGameScreen_Console.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+protected:
+	void _buttonClicked(Button* pButton) override;
+
+public:
+	void init() override;
+	void render(float f) override;
+	void renderPanel(float f) override;
+
+public:
```
BrentDaMage, line 32: Least accessible to most accessible.

source/client/gui/screens/LoadSaveScreen_Console.cpp
```diff
@@ -0,0 +1,70 @@
+#include "LoadSaveScreen_Console.hpp"
+#include "common/Util.hpp"
+#include "client/locale/Language.hpp"
+
+LoadSaveScreen_Console::LoadSaveScreen_Console(Options& options, Screen* parent, const LevelSummary& summary) :
```
BrentDaMage: Same as the others.

source/client/gui/screens/SavingWorldScreen_Console.cpp
```diff
@@ -0,0 +1,125 @@
...
+	m_bCopyMapAtEnd = bCopyMap;
+	m_timer = 0;
+	//m_pEntityToDeleteAfterSave = pEnt;
+	m_uiTheme = UI_CONSOLE;
+	Screen::init();
+
+	m_header = Language::get("saving.save");
+}
+
+void SavingWorldScreen_Console::render(float f)
```
BrentDaMage: Can this class inherit the typical loading screen, that way we don't have to have the render function copy-pasted?

source/client/gui/screens/SavingWorldScreen_Console.cpp
```diff
@@ -0,0 +1,125 @@
...
+	if (m_timer == 10)
+	{
+		m_pMinecraft->m_progressPercent = 66;
+		if (pLevel) pLevel->saveLevelData();
+	}
+
+	if (m_timer == 15)
+	{
+		m_pMinecraft->m_progressPercent = 100;
+		if (pLevel)
```
BrentDaMage: Please abstract this out into a function in `Minecraft`.

source/client/gui/screens/SavingWorldScreen_Console.cpp
```diff
@@ -0,0 +1,125 @@
...
+		m_pMinecraft->m_bUsingScreen = true;
+
+		if (m_bCopyMapAtEnd)
+			m_pMinecraft->setScreen(new RenameMPLevelScreen("_LastJoinedServer"));
+		else
+			m_pMinecraft->gotoMainMenu();
+
+		m_pMinecraft->m_bUsingScreen = false;
+
+		m_pMinecraft->m_bIsGamePaused = false;
```
BrentDaMage: Abstraction should end here.

source/client/gui/screens/SavingWorldScreen_Console.cpp
```diff
@@ -0,0 +1,125 @@
...
+		// this is safe to do, since on destruction, nothing accesses the parent level or anything
+		//SAFE_DELETE(m_pEntityToDeleteAfterSave);
+		// already done by the Level
+
+		m_pMinecraft->m_pCameraEntity = m_pMinecraft->m_pLocalPlayer = nullptr;
+
+
+		m_pMinecraft->m_bUsingScreen = true;
+
+		if (m_bCopyMapAtEnd)
```
BrentDaMage: This should just be passed as an argument.

source/client/gui/screens/SavingWorldScreen_Console.hpp
```diff
@@ -0,0 +1,29 @@
...
+
+class SavingWorldScreen_Console : public Screen
+{
+public:
+	SavingWorldScreen_Console(bool bCopyMap/*, Entity* pEntityToDeleteAfterSave*/);
+
+	void render(float f) override;
+	void tick() override;
+
+public:
```
BrentDaMage, line 18: Put in order of accessibility.

source/client/gui/screens/LoadSaveScreen_Console.hpp
```diff
@@ -0,0 +1,28 @@
+#pragma once
+
+#include "PanelScreen_Console.hpp"
+#include "client/gui/components/Button.hpp"
+#include "client/gui/components/TickBox.hpp"
+#include "client/gui/components/SliderButton.hpp"
+#include "client/gui/components/TextBox.hpp"
+
+class LoadSaveScreen_Console : public PanelScreen_Console
```
BrentDaMage: Call this `LoadWorldScreen_Console`

source/client/gui/Gui.cpp
```diff
@@ -557,10 +557,10 @@ void Gui::renderMessages(bool bShowAll)
 
 		if (m_pMinecraft->getOptions()->getUiTheme() == UI_CONSOLE)
 		{
-			fill(0, topEdge, GuiWidth, topEdge + 16 * scale, bkgdColor);
-			m_pMinecraft->m_pFont->drawScalableShadow(msg.msg, 30 * scale, topEdge + 5 * scale, textColor, scale);
+			fill(0, topEdge, GuiWidth, topEdge + 17 * scale, bkgdColor);
+			m_pMinecraft->m_pFont->drawScalable(msg.msg, 35 * scale, topEdge + 3 * scale, textColor, scale);
```
BrentDaMage, line 561: Comment out the old positioning instead of removing it completely.

source/client/gui/Screen.cpp
```diff
@@ -900,7 +900,10 @@ void Screen::renderBackground(int vo)
 	{
 		// draw the background offset by the Y offset so that the smaller virtual
 		// keyboards don't reveal undrawn areas
-		fillGradient(0, m_yOffset, m_width, m_height, Color(16, 16, 16, 192), Color(16, 16, 16, 208)); // 0xC0101010, 0xD0101010
+		if (m_uiTheme == UI_CONSOLE)
+			fillGradient(0, m_yOffset, m_width, m_height, Color(16, 16, 16, 100), Color(16, 16, 16, 112)); // 0x64101010, 0x70101010
```
BrentDaMage: You don't need to keep the hex version of this here, that's just there because I converted it from hex originally when I added the `Color` class alongside the HAL renderer.

source/client/gui/ScreenChooser_Console.hpp
```diff
@@ -19,4 +19,5 @@ class ScreenChooser_Console : public ScreenChooser
     void pushCraftingScreen(Player*, const TilePos&) override;
     void pushTrapScreen(Player*, DispenserTileEntity*) override;
     void pushDeathScreen() override;
+    virtual void pushSavingScreen(bool) override;
```
BrentDaMage: Remove the `virtual` keyword here.

source/client/renderer/entity/ItemRenderer.cpp
```diff
@@ -286,7 +286,7 @@ void ItemRenderer::renderGuiItem(Minecraft& mc, const ItemStack& item, int x, in
 			Lighting::turnOnConsoleUiItems();
 			matrix->translate(Vec3(x, y, 0));
 			matrix->scale(16);
-			matrix->translate(Vec3(0.5f, 0.5f, 0.0f));
+			matrix->translate(Vec3(0.53f, 0.53f, 0.0f));
```
BrentDaMage: I feel like we're just introducing scaling issues now. We're probably gonna have to come back thru a lot of this and fix the weird numbers in order to support split-screen.

source/client/renderer/entity/MobRenderer.cpp
```diff
@@ -224,6 +227,36 @@ void MobRenderer::renderNameTag(const Mob& mob, const std::string& str, const Ve
 	t.vertex(widthHalf + 1.0f, -1.0f, 0.0f);
 	t.draw(m_materials.name_tag);
 
+	currentShaderColor = Color(0.0f, 1.0f, 0.0f, 1.0f); //@TODO: Currently hardcoded to green, but should be changed to use different colors for players like in Xbox 360 Edition.
+
+#if MCE_GFX_API_OGL && !defined(FEATURE_GFX_SHADERS)
```
BrentDaMage: You're setting a line width regardless of the UI theme.

source/client/renderer/entity/MobRenderer.cpp
```diff
@@ -224,6 +227,36 @@ void MobRenderer::renderNameTag(const Mob& mob, const std::string& str, const Ve
 	t.vertex(widthHalf + 1.0f, -1.0f, 0.0f);
 	t.draw(m_materials.name_tag);
 
+	currentShaderColor = Color(0.0f, 1.0f, 0.0f, 1.0f); //@TODO: Currently hardcoded to green, but should be changed to use different colors for players like in Xbox 360 Edition.
```
BrentDaMage: You're setting the current color regardless of whether or not the outline should actually be drawn.

source/client/renderer/GameRenderer.cpp
```diff
@@ -224,7 +224,15 @@ void GameRenderer::_renderDebugOverlay(float a)
 	debugText << "\nGameRenderer::m_turnDelta.x: "      << m_turnDelta.x;
 	debugText << "\nGameRenderer::m_turnDelta.y: "      << m_turnDelta.y;*/
 
-	font.drawShadow(debugText.str(), 2, 2, Color::WHITE);
+	if (m_pMinecraft->getUiTheme() == UI_CONSOLE)
+	{
+		font.drawScalable(debugText.str(), 4, 4, Color(0.25f, 0.25f, 0.25f));
```
BrentDaMage: Isn't there a `Color::TEXT_SHADOW` or something like that?

source/client/renderer/ItemInHandRenderer.cpp
```diff
@@ -179,6 +179,9 @@ void ItemInHandRenderer::renderItem(const Entity& entity, const ItemStack& item,
 
         MatrixStack::Ref matrix = MatrixStack::World.push();
 
+        if (m_pMinecraft->getUiTheme() == UI_CONSOLE)
+            Lighting::turnOff();
```
BrentDaMage: Why are we only doing this on Console mode?

source/client/renderer/ScreenRenderer.cpp
```diff
@@ -25,6 +25,10 @@ std::string ScreenRenderer::POINTER_TEXT_PANEL_SLICES[] = { "gui/console/Graphic
                                                             "gui/console/Graphics/PointerTextPanel_ML.png", "gui/console/Graphics/PointerTextPanel_MM.png", "gui/console/Graphics/PointerTextPanel_MR.png",
                                                             "gui/console/Graphics/PointerTextPanel_BL.png", "gui/console/Graphics/PointerTextPanel_BM.png", "gui/console/Graphics/PointerTextPanel_BR.png" };
 
+std::string ScreenRenderer::PANEL_SQUARE_RECESS_SLICES[] = { "gui/console/Graphics/Square_Recess_Top_L.png", "gui/console/Graphics/Square_Recess_Top_M.png", "gui/console/Graphics/Square_Recess_Top_R.png",
+                                                      "gui/console/Graphics/Square_Recess_Mid_L.png", "gui/console/Graphics/Square_Recess_Mid_M.png", "gui/console/Graphics/Square_Recess_Mid_R.png",
```
BrentDaMage: Make this match the indentation of the others.

BrentDaMage (changes requested):
Of the full-res screenshots on the wiki, it does not look like the text shadows on buttons were ever like this. They all appear to be down and right by 1px on each axis.

source/client/renderer/Font.cpp
```diff
@@ -42,6 +42,8 @@ void Font::init(Options* pOpts)
 		if (i == 32) // space
 		{
 			widthMax = 2;
+			if (m_pOptions->getUiTheme() == UI_CONSOLE)
+				widthMax = 4;
```
BrentDaMage: We cannot afford to do this globally, since this could mess up Java screens that may appear while on the Console UI theme, particularly in the context of mods.

BrentDaMage:
Also, have you tested the nametag outline with shaders (either on OpenGL + Shaders, D3D9 or D3D1) to make sure it works correctly?

source/client/gui/screens/SavingWorldScreen_Console.cpp
fileshredder: Abstracted into Minecraft::unloadLevel

source/client/gui/Gui.cpp
fileshredder: But the old positioning does not match any title update and was a just a guess of the positions by me before i compared them correctly

source/client/renderer/entity/ItemRenderer.cpp
fileshredder: If it's an issue I could revert it

source/client/renderer/GameRenderer.cpp
fileshredder: replaced it with Color::TEXT_GREY

source/client/renderer/ItemInHandRenderer.cpp
fileshredder: Because in TU0/1/2/3/4, 4J accidentally broke normal lighting on items but kept it working on blocks, meaning items like swords, saplings and etc when held did not have any shading at all.

source/client/renderer/Font.cpp
fileshredder: Commented it out.

fileshredder:
> Also, have you tested the nametag outline with shaders (either on OpenGL + Shaders, D3D9 or D3D1) to make sure it works correctly?

I have not, but I could probably give it a try in a bit

fileshredder:
> > Also, have you tested the nametag outline with shaders (either on OpenGL + Shaders, D3D9 or D3D1) to make sure it works correctly?
> 
> I have not, but I could probably give it a try in a bit

tested it and it seems to work fine

source/client/gui/Gui.cpp
BrentDaMage: Alright then.

source/client/renderer/entity/ItemRenderer.cpp
BrentDaMage: I will test on different screen sizes and let you know how it looks.

source/client/renderer/ItemInHandRenderer.cpp
BrentDaMage: Do we *really* need to be doing this? I feel like most people would notice that it's broken before they realize that it was actually like that on older console versions.

fileshredder: yeah you're right, I'll revert it then.

source/client/gui/screens/AutosaveWarningScreen_Console.cpp
```diff
@@ -44,7 +45,7 @@ void AutosaveWarningScreen_Console::render(float f)
 	PanelScreen_Console::render(f);
 	Font& font = *m_pFont;
 
-	int yOffs = std::sin(getTimeMs() / 210.0f) * 4.0f;
+	int yOffs = std::sin((m_timer + f)/ 4.0f) * 4.0f;
```
BrentDaMage: Slight issue with this, is that you're relying on the `tick` function being called to increment the timer, which means that any FPS drops or missed ticks will result in inconsistent timing. While this isn't as much of an issue in this specific case, it may be better to use the Timer inside the Minecraft class, and track the time when the screen opens. Then, just calculate the time difference on each tick using subtraction, that way your timing is guaranteed to be accurate across all hardware.
See `ContainerScreen::handlePointerPressed()` for an example of similar behavior.

source/client/gui/screens/AutosaveWarningScreen_Console.hpp
```diff
@@ -2,18 +2,20 @@
...
 #include "client/gui/components/Button.hpp"
-#include "client/gui/components/TickBox.hpp"
-#include "client/gui/components/SliderButton.hpp"
-#include "client/gui/components/TextBox.hpp"
 
 class AutosaveWarningScreen_Console : public PanelScreen_Console
 {
-public:
-	AutosaveWarningScreen_Console(Screen*);
+private:
```
BrentDaMage: Member variables should always go at the bottom of the class. The variables themselves should be in order of least-accessible to most-accessible.

source/client/gui/screens/JoinGameScreen_Console.hpp
```diff
@@ -16,12 +14,16 @@ class UsernameButton : public Button
 
 class JoinGameScreen_Console : public PanelScreen_Console
 {
-public:
-	JoinGameScreen_Console(Screen*, const PingedCompatibleServer& server);
+private:
```
BrentDaMage: Same here.

source/client/gui/screens/LoadWorldScreen_Console.hpp
```diff
@@ -4,16 +4,15 @@
...
-class LoadSaveScreen_Console : public PanelScreen_Console
+class LoadWorldScreen_Console : public PanelScreen_Console
 {
-public:
-	LoadSaveScreen_Console(Options& options, Screen* parent, const LevelSummary& summary);
-
 protected:
 	void _buttonClicked(Button* pButton) override;
 
+public:
```
BrentDaMage, line 10: Look at the other files for reference on how things should be ordered.

source/client/gui/screens/ProgressScreen_Console.cpp
```diff
@@ -23,7 +24,7 @@ void ProgressScreen_Console::render(float f)
 
 	int loadingBarX = m_width / 2 - 320;
 	int loadingBarY = m_height / 2 + 30;
-	m_pFont->drawScalableShadow(m_pMinecraft->getProgressMessage(), loadingBarX + 7, loadingBarY - 15, Color::WHITE, 1.5f);
+	m_pFont->drawScalableShadow(m_subtext, loadingBarX + 7, loadingBarY - 15, Color::WHITE, 1.5f);
```
BrentDaMage, line 27: What if the result of `getProgressMessage()` changes while the screen is open?
If we determine the progress percent from `m_pMinecraft->m_progressPercent`, why would the text also come directly from `Minecraft`?

source/client/gui/screens/ProgressScreen_Console.hpp
```diff
@@ -9,10 +9,10 @@ class ProgressScreen_Console : public Screen
 	void updateEvents() override;
 	bool isInGameScreen() override;
 	void init() override;
-
-private:
+public:
```
BrentDaMage: Private member variables should come before public member variables.

source/client/gui/screens/SavingWorldScreen_Console.cpp
```diff
@@ -77,49 +30,27 @@ void SavingWorldScreen_Console::tick()
 	if (m_timer == 5)
 	{
 		m_pMinecraft->m_progressPercent = 33;
-		if (pLevel) pLevel->saveUnsavedChunks();
```
BrentDaMage: By moving these calls to `Minecraft`, we're completely losing the ability to track save progress. Don't do this.

BrentDaMage: Instead, you should set the progress percent inside of `Minecraft::unloadLevel()` before each function call, like what was done here. However, do not wait an arbitrary amount of time between each call, like what was done here.

source/client/gui/screens/SavingWorldScreen_Console.cpp
```diff
@@ -77,49 +30,27 @@ void SavingWorldScreen_Console::tick()
...
-		m_pMinecraft->m_bUsingScreen = false;
-
-		m_pMinecraft->m_bIsGamePaused = false;
+		m_pMinecraft->unloadLevel(m_bCopyMapAtEnd);
 
 		m_timer = -1;
 	}
 }
 
+void SavingWorldScreen_Console::updateEvents()
```
BrentDaMage, line 38: ?

source/client/gui/screens/SavingWorldScreen_Console.hpp
```diff
@@ -1,29 +1,28 @@
...
+#include "ProgressScreen_Console.hpp"
 
 #ifdef ENH_IMPROVED_SAVING
 
 class Entity;
 
-class SavingWorldScreen_Console : public Screen
+class SavingWorldScreen_Console : public ProgressScreen_Console
 {
+private:
```
BrentDaMage: Move this back down with all of the other member variables.

source/client/renderer/entity/MobRenderer.cpp
```diff
@@ -212,41 +210,26 @@ void MobRenderer::renderNameTag(const Mob& mob, const std::string& str, const Ve
...
 	t.normal(Vec3::UNIT_Y);
 	t.vertex(-1.0f - widthHalf, -1.0f, 0.0f);
 	t.vertex(-1.0f - widthHalf, 8.0f, 0.0f);
 	t.vertex(widthHalf + 1.0f, 8.0f, 0.0f);
 	t.vertex(widthHalf + 1.0f, -1.0f, 0.0f);
 	t.draw(m_materials.name_tag);
 
-	currentShaderColor = Color(0.0f, 1.0f, 0.0f, 1.0f); //@TODO: Currently hardcoded to green, but should be changed to use different colors for players like in Xbox 360 Edition.
-
-#if MCE_GFX_API_OGL && !defined(FEATURE_GFX_SHADERS)
```
BrentDaMage: Can you put an `@TODO` comment here to remind me to come back and do this later? I'm planning on adding something to the HAL for setting the line width, since this exists in OpenGL and 360 D3D9.

BrentDaMage (changes requested):

source/client/renderer/entity/MobRenderer.cpp
fileshredder: Are you sure? considering I only removed those lines since they werent needed or being used

BrentDaMage: Yes please. Technically the line thickness and smoothness may still work here, likely because we're never setting its state back after drawing the hitbox around blocks.

source/client/gui/screens/SavingWorldScreen_Console.cpp
fileshredder: I have to override this because otherwise the if statement that checks if the level is generated, from progressscreen_console will run, which will cause a softlock.

source/client/gui/screens/AutosaveWarningScreen_Console.cpp
fileshredder: Fixed it by now instead taking a snapshot of m_pMinecraft->m_timer.m_lastUpdateTime and storing it in m_timer and then doing the subtraction stuff.

source/client/gui/screens/ProgressScreen_Console.cpp
fileshredder: Fixed by making a new function in ProgressScreen_Console called updateText that returns the progressmessage and is overriden in SavingWorldScreen_console to instead return it's respective localization text.

source/client/gui/screens/SavingWorldScreen_Console.cpp
fileshredder: Done.

source/client/renderer/entity/MobRenderer.cpp
fileshredder: Done

source/client/gui/screens/SavingWorldScreen_Console.cpp
BrentDaMage: ah

source/client/gui/screens/AutosaveWarningScreen_Console.cpp
```diff
@@ -2,7 +2,7 @@
 #include "common/Util.hpp"
 #include "client/locale/Language.hpp"
 
-#define C_AUTOSAVE_TIMER (200)
+#define C_AUTOSAVE_TIMER (10)
```
BrentDaMage: What unit is this? Please put it in a comment or something.

source/client/gui/screens/AutosaveWarningScreen_Console.hpp
```diff
@@ -22,6 +14,14 @@ class AutosaveWarningScreen_Console : public PanelScreen_Console
 	void renderPanel(float f) override;
 	void tick() override;
 
+protected:
```
BrentDaMage: protected functions should go above public functions, but below the ctor

source/client/gui/screens/JoinGameScreen_Console.hpp
```diff
@@ -29,6 +22,13 @@ class JoinGameScreen_Console : public PanelScreen_Console
 	void render(float f) override;
 	void renderPanel(float f) override;
 
+protected:
```
BrentDaMage: should go above public functions but below the ctor

source/client/gui/screens/LoadWorldScreen_Console.hpp
```diff
@@ -18,6 +15,9 @@ class LoadWorldScreen_Console : public PanelScreen_Console
 	void render(float f) override;
 	void renderPanel(float f) override;
 
+protected:
```
BrentDaMage: same

source/client/gui/screens/ProgressScreen_Console.cpp
```diff
@@ -76,3 +75,8 @@ void ProgressScreen_Console::updateEvents()
 
 	Screen::updateEvents();
 }
+
+std::string ProgressScreen_Console::updateText()
```
BrentDaMage: call this `getSubText` and make it a const function

source/client/gui/screens/SavingWorldScreen_Console.cpp
```diff
@@ -54,3 +40,8 @@ void SavingWorldScreen_Console::updateEvents()
 }
 
 #endif
+
+std::string SavingWorldScreen_Console::updateText()
+{
+	return Language::get("saving.progress");
```
BrentDaMage: There's no reason for us to be spamming this call. Instead, we should have an underlying protected `m_subtext` variable in `ProgressScreen_Console`. Then, `getSubText()` should return this by default, but can be overridden if necessary.

BrentDaMage: Ideally we should not be calling Language::get every frame either.

source/client/gui/screens/ProgressScreen_Console.cpp
BrentDaMage: Also make it protected and prefix it with `_`

BrentDaMage (changes requested):

source/client/gui/screens/SavingWorldScreen_Console.cpp
fileshredder: made it return m_subtext instead and in progressscreen the text is updated in updateevents but i overrode it in savingworldscreen to not update it at all since we dont update sub text in that screen

source/client/gui/screens/JoinGameScreen_Console.hpp
```diff
@@ -16,15 +16,14 @@ class JoinGameScreen_Console : public PanelScreen_Console
 {
 public:
 	JoinGameScreen_Console(Screen*, const PingedCompatibleServer& server);
+protected:
```
BrentDaMage, line 20: put a new line above the accessibility

source/client/gui/screens/ProgressScreen_Console.hpp
```diff
@@ -9,12 +9,11 @@ class ProgressScreen_Console : public Screen
 	void updateEvents() override;
 	bool isInGameScreen() override;
 	void init() override;
-public:
-	virtual std::string updateText();
 private:
```
BrentDaMage, line 13: new line above accessibility specifier

source/client/gui/screens/SavingWorldScreen_Console.hpp
```diff
@@ -14,10 +14,6 @@ class SavingWorldScreen_Console : public ProgressScreen_Console
 	void tick() override;
 	void updateEvents() override;
 	void init() override;
-	std::string updateText() override;
-private:
-	std::string m_loadingTip;
-	std::vector<std::string> m_loadingTipLines;
 public:
```
BrentDaMage, line 18: new line above

source/client/gui/screens/SavingWorldScreen_Console.cpp
BrentDaMage: Everything looks good here post-changes.

source/client/renderer/entity/ItemRenderer.cpp
BrentDaMage: I just realized this is only done for 3D tiles, so scaling shouldn't be an issue.

BrentDaMage: Actually, all this does is misalign the 3D tiles in the inventory slightly to the right. Basically just making them off-center. I'm going to revert it, as I see zero reason to do this.

Image

BrentDaMage (approved):

BrentDaMage:
I will do a 360 build for QA tomorrow, then draft a changelog in our new format and pull upstream. The changelog will have "New features", "Changes", and "Bugfixes" sections like the last.

PR #683, opened 2026-08-01
Cleanup + bugfixes

### LINKED ISSUE #654: Tall grass has the dead bush texture
When did that happen?

BrentDaMage:
Closes #654


source/world/tile/TallGrass.cpp
```diff
@@ -41,7 +41,7 @@ Color TallGrass::getColor(TileSource& source, const TilePos& pos) const
 
 Color TallGrass::getColor(Facing::Name face, TileData data) const
 {
-	return data == 0 ? Color::WHITE : FoliageColor::getDefaultColor();
+	return data == 0 ? Color::WHITE : GrassColor::get(1.0f, 0.5f); // @PARITY-JAVA: should be 0xFF7CBD6B on b1.8, before that, nothing;
```
Un1q32, line 44: You could use that compile time version thing for this

BrentDaMage: I like our colors better anyways, so eh

PR #701, opened 2026-08-13
Reduced std::stringstream uses

### LINKED ISSUE #697: Reduce std::stringstream usage
Way more expensive than std::string concat or sprintf for no good reason

BrentDaMage:


source/client/options/Options.hpp
```diff
@@ -175,7 +175,7 @@ class OptionEntry
 	virtual const std::string& getName() const { return m_name; }
 	virtual const std::string& getDisplayName() const;
 	virtual std::string getDisplayValue() const;
-	virtual void save(std::stringstream&) const = 0;
+	virtual void save(std::string&) const = 0;
```
bluepilledgreat, line 178: Would be better to keep this as a stringstream imo

source/client/renderer/GameRenderer.cpp
```diff
@@ -186,9 +186,8 @@ void GameRenderer::_renderDebugOverlay(float a)
 {
 	Font& font = *m_pMinecraft->m_pFont;
 
-	std::stringstream debugText;
-	debugText << C_GAME_NAME " " << m_pMinecraft->getVersionString();
-	debugText << " (" << m_shownFPS << " fps, " << m_shownChunkUpdates << " chunk updates)" << "\n";
+	std::string debugText = C_GAME_NAME " " + m_pMinecraft->getVersionString();
```
bluepilledgreat, line 189: Ideas for the debug text:
- Move debug text rendering to a new class (DebugInfoRenderer)
- On enable, create a pointer to a string and reuse the same string to stop constant allocations
- On disable, nuke the string

source/common/Util.cpp
```diff
@@ -191,40 +191,76 @@ std::string Util::getExtension(const std::string& path)
...
-		absValue /= 10;
-	}
+#define _TOSTRING_S(_type, _digits)                                                 \
+std::string Util::toString(_type value)                                             \
+{                                                                                   \
+	if (value == 0)                                                                 \
+		return "0";                                                                 \
+		                                                                            \
+	/* X digits, 1 for sign */                                                      \
+	char buffer[_digits + 1];                                                       \
```
bluepilledgreat, line 201: Assuming null byte is not needed?

source/common/Util.cpp
```diff
@@ -191,40 +191,76 @@ std::string Util::getExtension(const std::string& path)
...
+		*(--ptr) = '0' + (value % 10);                                              \
+		value /= 10;                                                                \
+	}                                                                               \
+                                                                                    \
+	return std::string(ptr, sizeof(buffer) - (ptr - buffer));                       \
+}  
+
+#define _TOSTRING_IMPL(_type, _digits) \
+        _TOSTRING_S   (_type, _digits) \
+        _TOSTRING_U   (u##_type, _digits) \
```
bluepilledgreat: Assuming maximum digits is the same across the signed and unsigned types is completely wrong
int64_t can have 19 digits max
uint64_t can have 20 digits max

Remove the _TOSTRING_IMPL macro entirely please

bluepilledgreat (changes requested):

source/common/Util.cpp
BrentDaMage: I'm not assuming that they're the same. `_TOSTRING_S` adds an additional space for the sign.
Also, do you really want to have the same 28-line-long function copy-pasted 8 times?

source/client/options/Options.hpp
BrentDaMage: If you check the logic that actually calls this, all it's doing is allocating a new stringstream for every setting it reads. I would agree with you if we were streaming directly out to the file, but we're not.

source/common/Util.cpp
bluepilledgreat: You aren't addressing what I'm talking about. The problem is with int64. Signed can go up to 19 digits and unsigned can go up to 20 digits. With unsigned, the buffer size is only 19, which is very problematic.

source/client/renderer/GameRenderer.cpp
BrentDaMage: Make this a GitHub issue. I don't want to have to add new files to all of the projects for this PR.

source/common/Util.cpp
BrentDaMage: How does it look now?

source/common/Util.cpp
BrentDaMage: From my testing, yes.

bluepilledgreat (approved):

PR #708, opened 2026-08-15
adjust two areas of setTile which were ignoring data/misusing data; fixes #707

### LINKED ISSUE #704: Clients of multiplayer servers don't properly display variations of blocks.
As a client of a multiplayer server, when placing down a block that is a variation of an existing block (ex. colored wool, wood types), upon updating they lose the visual of their variation, (ex. red wool into white wool, birch logs into oak logs). This does not affect the host's view of the blocks, the blocks that the world saves, nor block variations that were placed down before the client joined the game.

Image

Image

### LINKED ISSUE #707: TileData Multiplayer Issues
I can see that data is being passed for birch/spruce, but it does not get used whenever the client receives the PlaceBlockPacket.
Also, whenever a variant block is placed and the tile updates, data is set back to 0.

Side note: is mayPlace supposed to always fail after ChunkDataPacket is finished initializing the world? seems the tile is being created before PlaceBlockPacket is recv'd; possibly due to useItem logic

lucanero:
adjust two areas of setTile which were ignoring data/misusing data; fixes 707

### DIFF
--- a/source/client/network/ClientSideNetworkHandler.cpp
+++ b/source/client/network/ClientSideNetworkHandler.cpp
@@ -392,9 +392,7 @@ void ClientSideNetworkHandler::handle(const RakNet::RakNetGUID& rakGuid, MovePla
 void ClientSideNetworkHandler::handle(const RakNet::RakNetGUID& rakGuid, PlaceBlockPacket* pPlaceBlockPkt)
 {
 	puts_ignorable("PlaceBlockPacket");
-
 	if (!m_pLevel) return;
-
 	Player* pPlayer = (Player*)m_pLevel->getEntity(pPlaceBlockPkt->m_entityId);
 	if (!pPlayer)
 	{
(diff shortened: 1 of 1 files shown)


source/client/network/ClientSideNetworkHandler.cpp
```diff
@@ -392,9 +392,7 @@ void ClientSideNetworkHandler::handle(const RakNet::RakNetGUID& rakGuid, MovePla
 void ClientSideNetworkHandler::handle(const RakNet::RakNetGUID& rakGuid, PlaceBlockPacket* pPlaceBlockPkt)
 {
 	puts_ignorable("PlaceBlockPacket");
-
```
BrentDaMage, line 395: fix formatting here

BrentDaMage (changes requested):

source/client/network/ClientSideNetworkHandler.cpp
lucanero: reverted whitespace removal

BrentDaMage (approved):

PR #716, opened 2026-08-15
Fix Server/Client PlaceBlockPacket and RemoveBlockPacket Sequencing

### LINKED ISSUE #715: Placing/Removing Blocks Not Replicating Properly
Currently, whenever a player places or breaks a block, it will play locally. When in multiplayer, it will send a BlockUpdate before PlaceBlockPacket or RemoveBlockPacket is sent. This causes the tile to update BEFORE the place/remove packet is received, and causes no sound or particle to appear for other clients aside from the host. 

The server currently uses UseItemPacket to place blocks, with PlaceBlockPacket being entirely redundant from my testing. I don't believe PlaceBlockPacket needs a server handler.

lucanero:
Closes #715

Does the server need a PlaceBlockPacket handler? Seems like a vulnerability and redundancy since placement is handled in UseItemPacket. Seems to be a client-only packet

### DIFF
--- a/source/server/ServerSideNetworkHandler.cpp
+++ b/source/server/ServerSideNetworkHandler.cpp
@@ -513,39 +513,6 @@ void ServerSideNetworkHandler::handle(const RakNet::RakNetGUID& guid, MovePlayer
 	}
 }
 
-void ServerSideNetworkHandler::handle(const RakNet::RakNetGUID& guid, PlaceBlockPacket* packet)
-{
-	if (!m_pLevel) return;
-	VALIDATE_PLAYER_ACTION(packet->m_entityId);
-
-	TilePos pos = packet->m_pos;
-
-	printf_ignorable("PlaceBlockPacket @ %d, %d, %d", pos.x, pos.y, pos.z);
-
-	player.swing();
-
-	TileID tileId = Tile::TransformToValidBlockId(packet->m_tileTypeId);
-	Facing::Name face = (Facing::Name)packet->m_face;
-	TileData data = packet->m_data;
-
-	TileSource& tileSource = player.getTileSource();
-
-	if (!tileSource.mayPlace(tileId, pos, face, player, true))
-		return;
-
-	if (tileSource.setTileAndData(pos, FullTile(tileId, data)))
-	{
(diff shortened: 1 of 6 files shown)


source/server/ServerSideNetworkHandler.cpp
```diff
@@ -513,39 +513,6 @@ void ServerSideNetworkHandler::handle(const RakNet::RakNetGUID& guid, MovePlayer
 	}
 }
 
-void ServerSideNetworkHandler::handle(const RakNet::RakNetGUID& guid, PlaceBlockPacket* packet)
```
BrentDaMage, line 516: A possible reason to keep this:
Vanilla PE clients will place their block regardless, so if the item they're holding doesn't exist on the server, this function still ensures a placeholder block is placed to maintain sync.

source/world/gamemode/GameMode.cpp
```diff
@@ -45,22 +45,19 @@ bool GameMode::destroyBlock(Player& player, const TilePos& pos, Facing::Name fac
 
 	oldTile->playerWillDestroy(player, pos, face);
 
+	Level& level = player.getLevel();
+	if (m_pMinecraft->isOnline() && player.isLocalPlayer())
+		level.m_pRakNetInstance->send(new RemoveBlockPacket(player.m_EntityID, pos));
```
BrentDaMage: Why are we sending this regardless of whether or not the Tile changed?

source/world/item/TileItem.cpp
```diff
@@ -52,6 +52,9 @@ bool TileItem::useOn(ItemStack& itemStack, Player& player, const TilePos& pos, F
 	if (!source.mayPlace(m_tile, tp, face, player))
 		return false;
 
+	if (level.m_pRakNetInstance)
+		level.m_pRakNetInstance->send(new PlaceBlockPacket(player.m_EntityID, tp, (TileID)m_tile, face, itemStack.getAuxValue()));
```
BrentDaMage: Why are we sending this regardless of whether or not the block was successfully placed?

BrentDaMage (changes requested):

BrentDaMage (approved):

PR #724, opened 2026-08-19
Text rendering refactor, caching and Unicode support

bluepilledgreat:
Changes to text rendering:
- Refactored code
  - Magic values cleaned up
  - Proper variable names
- Fixed functionality for certain functions, including Font::height
- Unicode (plane 0 to 1) and UTF-8 support (with the aid of utf8proc), opening up the door for localisation (plus symbols and emojis)
- Text mesh caching to reduce the number of times a text's mesh has to be rebuilt
- Text formatting codes
- Removed `isConsole` arguments as console specific text rendering has been moved to a new class (ConsoleFont)

Text rendering has been split up into three different functions:
- drawCached: New primary font rendering. Features caching, Unicode, etc.
- drawSimple: Old font rendering. Only capable of rendering Ascii, without any caching. Should be used for rapidly changing data (e.g. debug menu).
- drawSimpleBatchable: Same as above, with the advantage of being able to be batched (for profiler specifically).

TODO:
- Font::split
  - Rename to Font::break
  - Make UTF-8 compliant
- Add drawSimpleBatchable
- Obfuscation text formatting (maybe for another PR?)
- ~~Fix warnings~~
- Update other project files and make files
- Commit glyph maps (excluded until confirmed to be final, ask me for maps if you wish to test)
- ~~Look into having toggles for disabling Unicode and UTF-8 support (up for discussion)~~


projects/visual-studio/Client/Client.vcxproj
```diff
@@ -76,6 +76,78 @@
     <Import Project="../Directory.Build.props" Condition="'$(MC_GLOBAL_PROPS)'==''" />
   </ImportGroup>
   <PropertyGroup Label="UserMacros" />
+  <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
```
BrentDaMage, line 79: don't pollute the project files

projects/visual-studio/Client/Client.vcxproj
```diff
@@ -518,6 +592,9 @@
       <Project>{f332cb57-ff16-4d43-bbe0-76f28301daa8}</Project>
       <LinkLibraryDependencies>true</LinkLibraryDependencies>
     </ProjectReference>
+    <ProjectReference Include="..\utf8proc\utf8proc.vcxproj">
```
BrentDaMage, line 595: Please use the same casing as other projects. ex: `UTF8Proc`

projects/visual-studio/utf8proc/utf8proc.vcxproj
```diff
@@ -0,0 +1,152 @@
+<?xml version="1.0" encoding="utf-8"?>
```
BrentDaMage, line 1: use the `Common` or `STB` projects as examples

projects/visual-studio/NBCraft.sln
```diff
@@ -1,4 +1,4 @@
-Microsoft Visual Studio Solution File, Format Version 11.00
+Microsoft Visual Studio Solution File, Format Version 12.00
```
BrentDaMage, line 1: no

source/client/app/Minecraft.cpp
```diff
@@ -1459,6 +1460,34 @@ UITheme Minecraft::getUiTheme()
...
 }
 
+Font* Minecraft::_createFont(const std::string& fileName) const
+{
+	assert(m_pOptions);
+
+	switch (m_pOptions->getUiTheme())
+	{
+	case UI_CONSOLE:
+		return new ConsoleFont(m_pOptions, fileName, m_pTextures);
```
BrentDaMage, line 1470: Call this `Font_Console`

source/client/app/Minecraft.cpp
```diff
@@ -1459,6 +1460,34 @@ UITheme Minecraft::getUiTheme()
...
+
+void Minecraft::_reloadFontInternal()
+{
+	delete m_pFont;
+	m_pFont = _createFont("font/default.png");
+}
+
+void Minecraft::reloadFont()
+{
+	if (!m_initialized)
```
BrentDaMage, line 1485: couldn't we just check if m_pFont is null?

source/client/app/Minecraft.hpp
```diff
@@ -119,6 +119,22 @@ class Minecraft : public App
...
+		return m_initialized;
+	}
+
+	void reloadFont();
+
+protected:
+	Font* _createFont(const std::string& fileName) const;
+	void _reloadFontInternal();
+
+	void markInitialized()
```
BrentDaMage, line 133: not a fan of introducing this just for Fonts

source/client/gui/screens/StartMenuScreen.cpp
```diff
@@ -166,7 +166,8 @@ void StartMenuScreen::init()
 
 	_addElement(m_creditsButton, false);
 
-	m_watermarkText = "\xFFMojang AB";
+	// TODO: change copyright symbol back to \xFF if utf8 is disabled
+	m_watermarkText = "\xC3\xBFMojang AB";
```
BrentDaMage, line 170: any chance we could do an ASCII draw here, since this can't really be localized?

source/client/renderer/Font.cpp
```diff
@@ -1,102 +1,839 @@
...
-
-#include "Font.hpp"
+#include "client/renderer/Font.hpp"
 #include "client/renderer/renderer/RenderMaterialGroup.hpp"
 #include "renderer/ShaderConstants.hpp"
-#include "renderer/MatrixStack.hpp"
+#include "common/utility/hashing/HashCombine.hpp"
+#include "common/Util.hpp"
 #include <sstream>
+#include <utf8proc.h>
```
BrentDaMage, line 7: Do `thirdparty/utf8proc/utf8proc.h` so we don't have to add this to the include paths on every IDE + CMake.

source/client/renderer/Font.cpp
```diff
@@ -1,102 +1,839 @@
...
- ********************************************************************/
-
-#include "Font.hpp"
+#include "client/renderer/Font.hpp"
 #include "client/renderer/renderer/RenderMaterialGroup.hpp"
 #include "renderer/ShaderConstants.hpp"
-#include "renderer/MatrixStack.hpp"
+#include "common/utility/hashing/HashCombine.hpp"
+#include "common/Util.hpp"
 #include <sstream>
```
BrentDaMage, line 6: also, STL includes should always come before project-specific includes

source/client/renderer/Font.cpp
```diff
@@ -1,102 +1,839 @@
...
+	Color::FromRGB(85, 85, 255),   // 9 - blue
+	Color::FromRGB(85, 255, 85),   // a - green
+	Color::FromRGB(85, 255, 255),  // b - aqua
+	Color::FromRGB(255, 85, 85),   // c - red
+	Color::FromRGB(255, 85, 255),  // d - light purple
+	Color::FromRGB(255, 255, 85),  // e - yellow
+	Color::FromRGB(255, 255, 255)  // f - white
+};
+
+constexpr int MAX_CACHE_SIZE = 500;
```
BrentDaMage, line 28: make all of these macros

source/client/renderer/Font.cpp
```diff
@@ -1,102 +1,839 @@
...
+	, italic(italic)
+	, isAscii(isAscii)
+{
+}
+
+void Font::GlyphQuad::append(Tesselator& t)
+{
+	t.color(color);
+
+	const int mapGlyphSize = isAscii ? ASCII_MAP_GLYPH_SIZE : UNICODE_MAP_GLYPH_SIZE;
```
BrentDaMage, line 73: we don't need to put const in front of everything here, afaik it makes no difference to the compiler

BrentDaMage: @Un1q32 can you confirm?

source/client/renderer/Font.cpp
BrentDaMage: also prefix them with `C_`

source/client/renderer/Font.cpp
```diff
@@ -1,102 +1,839 @@
...
+	// if this glyph is italic, shift the top corner by X to the right
+	float shift = italic ? ITALIC_SHIFT : 0.0f;
+
+	t.vertexUV(x,                             y + RENDER_GLYPH_SIZE, 0.0f, u * D,                  (v + mapGlyphSize) * D);
+	t.vertexUV(x + RENDER_GLYPH_SIZE,         y + RENDER_GLYPH_SIZE, 0.0f, (u + mapGlyphSize) * D, (v + mapGlyphSize) * D);
+	t.vertexUV(x + RENDER_GLYPH_SIZE + shift, y,                     0.0f, (u + mapGlyphSize) * D, v * D);
+	t.vertexUV(x + shift,                     y,                     0.0f, u * D,                  v * D);
+}
+
+Font::TextObject::TextObject()
```
BrentDaMage, line 90: pretty sure you don't need to even have these in the header

source/client/renderer/Font.cpp
```diff
@@ -1,102 +1,839 @@
...
+{
+	for (std::vector<Page>::iterator it = pages.begin(); it != pages.end(); it++)
+	{
+		Page& page = *it;
+		page.textureData->bind();
+		page.mesh.render(material);
+	}
+}
+
+Font::TextObject::Page::Page(mce::Mesh& mesh, TextureData* textureData)
```
BrentDaMage, line 114: any chance we could get this stuff in-lined?

source/client/renderer/Font.cpp
```diff
@@ -1,102 +1,839 @@
...
 	{
-		// note: the 'widthMax' behavior is assumed. It might not be like that exactly
-		int widthMax = 0;
+		if (!base)
+			base = font._createTextObject(str, color, false);
 
-		if (i == 32) // space
+		return *base;
+	}
+	else
```
BrentDaMage, line 154: remove the else

source/client/renderer/Font.cpp
```diff
@@ -1,102 +1,839 @@
...
+}
+
+void Font::Line::append(const Font& font, Tesselator& t, float height, float yShift)
+{
+	if (fromX == toX)
+		return;
+
+	assert(font.m_pixelX != -1 && font.m_pixelY != -1);
+
+	const int u = font.m_pixelX;
```
BrentDaMage, line 206: don't think const is necessary here

source/client/renderer/Font.cpp
```diff
@@ -1,102 +1,839 @@
...
+	const int v = font.m_pixelY;
+
+	const int mapSize = ASCII_MAP_PIXEL_DIMENSION; // always the ascii map for now
```
BrentDaMage, line 209: make these both constexpr

source/client/renderer/Font.cpp
```diff
@@ -1,102 +1,839 @@
...
+	, m_resetFormatOnBuild(true)
+	, m_pixelX(-1)
+	, m_pixelY(-1)
+	, m_unicodeShadowOffset(RENDER_GLYPH_SIZE / UNICODE_MAP_GLYPH_SIZE)
+{
+	m_recentTextObjectCaches.reserve(MAX_CACHE_SIZE);
+	_init(options);
+}
+
+Font::~Font()
```
BrentDaMage, line 276: this doesn't need to be declared or defined

BrentDaMage (changes requested):
partial review

projects/visual-studio/Client/Client.vcxproj
bluepilledgreat: That is the correct casing for this library, would prefer to keep that.

source/client/renderer/Font.cpp
bluepilledgreat: It should get inlined, only usage of this is inside this file

source/client/renderer/Font.cpp
bluepilledgreat: Does it really matter?

source/client/renderer/Font.cpp
bluepilledgreat: It does, I have ~Font defined as a virtual.

source/client/app/Minecraft.cpp
bluepilledgreat: When options are loaded, UIThemeOption::apply is called which results in a call to reloadFont. This is way before the font is normally loaded.

source/client/app/Minecraft.hpp
bluepilledgreat: See Minecraft::reloadFont comment

source/client/gui/screens/StartMenuScreen.cpp
bluepilledgreat: We preferably shouldn't as drawSimple does not cache. Caching introduces serious performance gains, even for small text such as that.

source/client/renderer/Font.cpp
bluepilledgreat: Very ugly but ok

source/client/renderer/Font.cpp
bluepilledgreat: Will be removed once Font::split is done

source/client/renderer/Font.cpp
bluepilledgreat: Does not really matter

source/client/renderer/Font.cpp
bluepilledgreat: I did at one point as I wasn't holding TextObject pointers at one point.
Does not really matter if they're in the headers or not.

projects/visual-studio/utf8proc/utf8proc.vcxproj
bluepilledgreat: dude i really dont want to create a whole ass visual studio project by hand if i'm going to do premake next

BrentDaMage: I don't care. I'm not letting you pollute the project structure with a gazillion copy-pasted settings.

bluepilledgreat: boohoo im doing premake right now im not dealing with this

projects/visual-studio/Client/Client.vcxproj
BrentDaMage: fffffiine

source/client/renderer/Font.cpp
BrentDaMage: ah

source/common/math/Color.hpp
```diff
@@ -79,6 +80,16 @@ struct Color
         return *this;
     }
 
+    uint32_t toUInt32() const
```
BrentDaMage, line 83: make this an explicit conversion operator

bluepilledgreat: This was discussed iirc
Not a good idea for code quality

